PR #3, opened 2020-09-08
Check error code rather than custom implement get_string

DoumanAsh:
JFYI: you can check error code if you want a better errors so I made example of get_string

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,12 +13,7 @@ edition = "2018"
 thiserror = "1.0"
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.9", features = [
-    "winuser",
-    "winbase",
-    "stringapiset",
-    "basetsd",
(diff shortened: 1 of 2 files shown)


ArturKovacs:
I appretiate your work on this.

It's a good idea to minimize re-implementing stuff that already exists; however I don't think this is a reliable way to handle errors. The problem is that the error codes `ERROR_INVALID_FLAGS` and `ERROR_INVALID_PARAMETER` are very generic and they can come from many different winapi functions. I understand that using the current implementation, this is correct, but the `clipboard_win::get(clipboard_win::Unicode)` function-call gives me no guarantees that this is not going to change in a future release. And if it does change, I get no warnings or errors, the code will just compile without me knowing that it's incorrect.

DoumanAsh:
Ok, this code has no need to change (as it is the only way to get unicode string) but your concern is valid

PR #5, opened 2020-10-25
Preserve clipboard contents on exit on GNOME 3.34 and later

### LINKED ISSUE #4: Text placed into the X11 clipboard not available afterwards
With `examples/hello_world.rs` and initial "some old contents" clipboard contents:

```
$ target/debug/examples/hello_world
Clipboard text was: some old content
But now the clipboard text should be: "Hello, world!"

$ target/debug/examples/hello_world
Clipboard text was: some old content
But now the clipboard text should be: "Hello, world!"

$ target/debug/examples/hello_world
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ContentNotAvailable - "The clipboard contents were not available in the requested format or the clipboard is empty."', examples/hello_world.rs:5:61
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The first run appears to have no effect on the X11 clipboard; the second one does have some effect, but only to clear it.

Note that the panic in the third run is only a side effect: the example in question doesn't handle the case where the clipboard is empty.

After some debugging, it seems that `x11_clipboard_manager` is 0 during the `destruct` call, [preventing the handing off of the data](https://github.com/ArturKovacs/arboard/blob/2f3fde214745c09ed609a4316e661a3dfcf68b15/src/x11_clipboard.rs#L544-L557).

I'm not sure why the `x11_clipboard_manager != 0` condition is necessary, and if fact it seems that a `xcb_window_t` value of 0 (or `XCB_NONE`) is valid for other XCB calls; for example, in `xcb_set_selection_owner` it [means that the selection will have no owner](https://www.x.org/releases/X11R7.7/doc/man/man3/xcb_set_selection_owner.3.xhtml).

I experimented with removing this condition entirely (and the previous calls that only serve to fill in `x11_clipboard_manager`) and everything worked just fine, but I'm well aware that this doesn't mean anything, particularly when dealing with Xorg. : )

P.S. I first noticed the problem in 1.0.1, but I've been since working on top of the current HEAD.

jonasmalacofilho:
Fixes #4

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,10 +102,15 @@ fn all_tests() {
 		ctx.set_text(text.to_owned()).unwrap();
 		assert_eq!(ctx.get_text().unwrap(), text);
 
+		// We also need to check that the content persists after the drop; this is
+		// especially important on X11
 		drop(ctx);
 
+		// Give any external mechanism a generous amount of time to take over
+		// responsibility for the clipboard, in case that happens asynchronously
+		// (it appears that this is the case on X11 plus Mutter 3.34+, see #4)
 		use std::time::Duration;
-		std::thread::sleep(Duration::from_millis(10));
+		std::thread::sleep(Duration::from_millis(50));
 
 		let mut ctx = Clipboard::new().unwrap();
 		assert_eq!(ctx.get_text().unwrap(), text);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,12 +93,21 @@ impl Clipboard {
 
 /// All tests grouped in one because the windows clipboard cannot be open on
 /// multiple threads at once.
+#[cfg(test)]
 #[test]
 fn all_tests() {
 	{
 		let mut ctx = Clipboard::new().unwrap();
 		let text = "some string";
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -93,13 +93,22 @@ impl Clipboard {
...
+#[cfg(test)]
 #[test]
 fn all_tests() {
 	{
 		let mut ctx = Clipboard::new().unwrap();
 		let text = "some string";
 		ctx.set_text(text.to_owned()).unwrap();
 		assert_eq!(ctx.get_text().unwrap(), text);
+
+		drop(ctx);
```
ArturKovacs, line 105: I think it would be nice to leave a comment explaining that this drop is needed because we test if the clipbard hands over the contents corretly to the clipboard manager on Linux.

ArturKovacs:
Feel free to change this from a draft to a normal PR. I'd be happy to merge it.

ArturKovacs:
I managed to get back to this a bit later than I wanted to, sorry for the delay.

ArturKovacs:
I updated your top comment so the related issue will get closed automatically.

PR #9, opened 2021-03-08
Replace `xcb` with `x11rb`

### LINKED ISSUE #8: Replace or add an alternative to `xcb` with `x11rb`
Considering that `xcb` has been marked by `cargo-audit` as problematic: https://rustsec.org/advisories/RUSTSEC-2021-0019.html
Would a PR to replace it with `x11rb` be accepted, or at least use it as an alternative with the help of a crate feature?

daxpedda:
I really only did a 1:1 translation. I believe `x11rb` can do much better here. ~~I am still very new to X11, currently experimenting with a better solution, but this is totally mergable.~~

- `libc` was dropped in the process
- Except porting this 1:1 I only did four very small changes:
  - Replaced our `Atom` storage with [`atom_manager!`](https://docs.rs/x11rb/0.8.0/x11rb/macro.atom_manager.html)
  - Instead of picking the first screen, pick the one returned by `connect`
  - use `COPY_DEPTH_FROM_PARENT` instead of 0
  - resolved any other issues Clippy pointed out
- There are a number of functions here that return an unhandled `Result`, seeing that these errors were unhandled before I left everything as it was, ~~I will fix it if I manage to improve things~~ should I just `unwrap` them?
- I had to apply `rustfmt` on an unrelated file to make CI run

Fixes #8.

### DIFF
--- a/src/x11_clipboard.rs
+++ b/src/x11_clipboard.rs
@@ -388,14 +388,14 @@ impl Manager {
 		match &mut self.reply_data {
 			None => {
 				self.reply_offset = 0; // Rust impl: I added this just to be extra sure.
-				self.reply_data = Some(Arc::new(Mutex::new(vec![0u8; req])));
+				self.reply_data = Some(Arc::new(Mutex::new(vec![0; req])));
 			}
 			// The "m_reply_data" size can be smaller because the size
 			// specified in INCR property is just a lower bound.
 			Some(reply_data) => {
 				let mut reply_data = reply_data.lock().unwrap();
 				if req > reply_data.len() {
-					reply_data.resize(req, 0u8);
+					reply_data.resize(req, 0);
 				}
 			}
 		}
@@ -460,10 +460,8 @@ impl Manager {
 			}
 
 			if manager!().window != 0 {
-				{
-					let guard = guard.as_mut().unwrap();
-					guard.shared.conn.as_ref().unwrap().destroy_window(guard.manager.window);
-				}
+				let window = manager!().window;
+				shared!().conn.as_ref().unwrap().destroy_window(window);
 				shared!().conn.as_ref().unwrap().flush();
 				manager!().window = 0;
 			}
--- a/src/x11_clipboard.rs
+++ b/src/x11_clipboard.rs
@@ -73,12 +73,12 @@ x11rb::atom_manager! {
 x11rb::atom_manager! {
 	pub TextAtoms: TextAtomCookies {
 		UTF8_STRING,
-		TEXT_PLAN_1: b"text/plain;charset=utf-8",
-		TEXT_PLAN_2: b"text/plain;charset=UTF-8",
+		TEXT_PLAIN_1: b"text/plain;charset=utf-8",
+		TEXT_PLAIN_2: b"text/plain;charset=UTF-8",
 		// ANSI C strings?
 		STRING,
 		TEXT,
-		TEXT_PLAN_0: b"text/plain",
+		TEXT_PLAIN_0: b"text/plain",
 	}
 }
 
@@ -149,11 +149,11 @@ impl SharedState {
 
 		vec![
 			atoms.UTF8_STRING,
-			atoms.TEXT_PLAN_1,
-			atoms.TEXT_PLAN_2,
+			atoms.TEXT_PLAIN_1,
+			atoms.TEXT_PLAIN_2,
 			atoms.STRING,
 			atoms.TEXT,
-			atoms.TEXT_PLAN_0,
+			atoms.TEXT_PLAIN_0,
 		]
 	}
 }
--- a/src/x11_clipboard.rs
+++ b/src/x11_clipboard.rs
@@ -38,19 +38,21 @@ use std::rc::Rc;
 use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use std::time::Duration;
 
-use image;
 use lazy_static::lazy_static;
-use libc;
-use xcb::ffi::base::{xcb_request_check, XCB_CURRENT_TIME};
-use xcb::ffi::xproto::{
-	xcb_atom_t, xcb_change_property, xcb_get_property, xcb_get_property_reply,
-	xcb_get_property_reply_t, xcb_get_property_value, xcb_get_property_value_length,
-	xcb_get_selection_owner_reply, xcb_property_notify_event_t, xcb_selection_clear_event_t,
-	xcb_selection_notify_event_t, xcb_selection_request_event_t, xcb_send_event,
-	xcb_set_selection_owner_checked, XCB_ATOM_NONE, XCB_EVENT_MASK_NO_EVENT,
-	XCB_PROPERTY_NEW_VALUE, XCB_PROP_MODE_REPLACE, XCB_SELECTION_NOTIFY,
+use x11rb::protocol::xproto;
+use x11rb::{
+	connection::Connection,
+	protocol::{
+		xproto::{
+			Atom, AtomEnum, ConnectionExt as _, CreateWindowAux, EventMask, GetPropertyReply,
+			PropMode, Property, PropertyNotifyEvent, SelectionClearEvent, SelectionNotifyEvent,
+			SelectionRequestEvent, Time, Window, WindowClass,
+		},
+		Event,
+	},
+	wrapper::ConnectionExt as _,
+	xcb_ffi::XCBConnection,
 };
-use xcb::xproto;
 
 use super::common::{Error, ImageData};
 
@@ -64,7 +66,7 @@ const SAVE_TARGETS: usize = 6;
 const MULTIPLE: usize = 7;
 const CLIPBOARD_MANAGER: usize = 8;
 
-static COMMON_ATOM_NAMES: [&'static str; 9] = [
+static COMMON_ATOM_NAMES: [&str; 9] = [
 	"ATOM",
 	"INCR",
 	"TARGETS",
@@ -77,7 +79,7 @@ static COMMON_ATOM_NAMES: [&'static str; 9] = [
 ];
 
 type BufferPtr = Option<Arc<Mutex<Vec<u8>>>>;
-type Atoms = Vec<xcb::xproto::Atom>;
+type Atoms = Vec<Atom>;
 type NotifyCallback = Option<Arc<dyn (Fn(&BufferPtr) -> bool) + Send + Sync + 'static>>;
 
 lazy_static! {
@@ -95,8 +97,8 @@ struct LockedObjects {
 
 impl LockedObjects {
 	fn new() -> Result<LockedObjects, Error> {
-		let connection = xcb::Connection::connect(None).unwrap().0;
-		match Manager::new(&connection) {
+		let (connection, screen) = XCBConnection::connect(None).unwrap();
+		match Manager::new(&connection, screen) {
 			Ok(manager) => {
 				//unsafe { libc::atexit(Manager::destruct); }
 				Ok(LockedObjects {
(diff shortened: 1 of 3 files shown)


src/x11_clipboard.rs
```diff
@@ -273,45 +273,33 @@ struct Manager {
...
-		}
-		let screen = setup.roots().data;
-		if std::ptr::null() == screen {
-			return Err(Error::Unknown { description: "Could not get screen from setup".into() });
-		}
+	fn new(connection: &XCBConnection, screen: usize) -> Result<Self, Error> {
+		let setup = connection.setup();
+		let screen = setup.roots.get(screen).ok_or(Error::Unknown {
+			description: String::from("Could not get screen from setup"),
+		})?;
```
daxpedda, line 280: Instead of gettings the first screen, we are getting the correct screen passed down from `connect`.

src/x11_clipboard.rs
```diff
@@ -273,45 +273,33 @@ struct Manager {
...
-				XCB_CW_EVENT_MASK,
-				&event_mask,
-			);
-		}
+            EventMask::STRUCTURE_NOTIFY;
+		let window = connection.generate_id().map_err(|_| Error::Unknown {
+			description: String::from("Could not generate ID from connection"),
+		})?;
+		connection.create_window(
+			x11rb::COPY_DEPTH_FROM_PARENT,
```
daxpedda, line 291: Instead of `0` depth, we copy from parent.

daxpedda:
@ArturKovacs I don't plan to improve this any further unless necessary.
Long-term I believe we should try to adopt this code: https://github.com/hecrj/window_clipboard/blob/master/x11/src/lib.rs

src/x11_clipboard.rs
ArturKovacs: I'm wondering why is this better, when the window remains hidden, and is never drawn on.

src/x11_clipboard.rs
```diff
@@ -33,51 +33,57 @@ and conditions of the chosen license apply to this file.
...
+		INCR,
+		TARGETS,
+		CLIPBOARD,
+		MIME_IMAGE_PNG: b"image/png",
+		ATOM_PAIR,
+		SAVE_TARGETS,
+		MULTIPLE,
+		CLIPBOARD_MANAGER,
+	}
+}
```
ArturKovacs, line 71: This is awesome!

src/x11_clipboard.rs
```diff
@@ -33,51 +33,57 @@ and conditions of the chosen license apply to this file.
...
 use std::cell::RefCell;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::rc::Rc;
 use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use std::time::Duration;
 
-use image;
 use lazy_static::lazy_static;
-use libc;
```
ArturKovacs, line 43: No `libc`, yaay!

src/x11_clipboard.rs
```diff
@@ -273,45 +227,33 @@ struct Manager {
...
+			screen.root,
+			0,
+			0,
+			1,
+			1,
+			0,
+			WindowClass::INPUT_OUTPUT,
+			screen.root_visual,
+			&CreateWindowAux::new().event_mask(event_mask),
+		);
```
ArturKovacs, line 256: This error should be propagated by something like this:
```rust
).map_err(|e| Error::Unknown { description: format!("{}", e) })?;
```

src/x11_clipboard.rs
```diff
@@ -437,49 +369,44 @@ impl Manager {
...
-				item.as_ptr() as *const _,
-			)
-		};
+		shared.conn.as_ref().unwrap().change_property8(
+			PropMode::REPLACE,
+			requestor,
+			property,
+			target,
+			item.as_slice(),
+		);
```
ArturKovacs, line 378: Let's bring in the `log` crate and use either `log::error!` or `log::warn!` here, to log the error message, but leave it at that. I don't know what it means for us if this fails so I leave it up to you to decide which one you use.

src/x11_clipboard.rs
```diff
@@ -533,12 +460,8 @@ impl Manager {
...
 			if manager!().window != 0 {
-				unsafe {
-					xcb::ffi::xproto::xcb_destroy_window(
-						shared!().conn.as_ref().unwrap().get_raw_conn(),
-						manager!().window,
-					)
-				};
+				let window = manager!().window;
+				shared!().conn.as_ref().unwrap().destroy_window(window);
 				shared!().conn.as_ref().unwrap().flush();
```
ArturKovacs, line 465: These should be ignored. (Definitely *not* unwrapped because this is called from a `drop`, and we probably don't want to panic while potentially another panic is happening)
```suggestion
				let _ = shared!().conn.as_ref().unwrap().destroy_window(window);
				let _ = shared!().conn.as_ref().unwrap().flush();
```

src/x11_clipboard.rs
```diff
@@ -673,25 +578,20 @@ fn handle_selection_request_event(event: *mut xcb_selection_request_event_t) {
...
-				XCB_PROP_MODE_REPLACE as u8,
-				requestor,
-				property,
-				atom_atom,
-				8 * std::mem::size_of::<xcb_atom_t>() as u8,
-				targets.len() as u32,
-				targets.as_ptr() as *const _,
-			)
-		};
+		shared.conn.as_ref().unwrap().change_property32(
```
ArturKovacs, line 581: Just like before, let's use either `log::error!` or `log::warn!` here.

src/x11_clipboard.rs
```diff
@@ -700,48 +600,29 @@ fn handle_selection_request_event(event: *mut xcb_selection_request_event_t) {
...
-					};
+					locked.shared.conn.as_ref().unwrap().change_property(
+						PropMode::REPLACE,
+						requestor,
+						property,
+						AtomEnum::NONE,
+						0,
+						0,
+						&[],
+					);
```
ArturKovacs, line 623: Use `log::error!` or `log::warn!` here as well.

src/x11_clipboard.rs
```diff
@@ -762,112 +643,83 @@ fn handle_selection_request_event(event: *mut xcb_selection_request_event_t) {
...
-		xcb_send_event(
-			shared.conn.as_ref().unwrap().get_raw_conn(),
-			0,
-			requestor,
-			XCB_EVENT_MASK_NO_EVENT,
-			&notify as *const _ as *const _,
-		)
-	};
+	shared.conn.as_ref().unwrap().send_event(false, requestor, EventMask::NO_EVENT, notify);
 	shared.conn.as_ref().unwrap().flush();
```
ArturKovacs, line 656: Just as before, `log::error!` or `log::warn!`.

src/x11_clipboard.rs
```diff
@@ -952,14 +791,13 @@ fn get_data_from_selection_owner<'a>(
...
-				locked.shared.conn.as_ref().unwrap(),
+			let clipboard_atom = locked.shared.common_atoms().CLIPBOARD;
+			locked.shared.conn.as_ref().unwrap().convert_selection(
 				locked.manager.window,
 				selection,
 				*atom,
 				clipboard_atom,
-				xcb::base::CURRENT_TIME,
+				Time::CURRENT_TIME,
 			);
```
ArturKovacs, line 801: `log:error!` or `log::warn!`

src/x11_clipboard.rs
```diff
@@ -952,14 +791,13 @@ fn get_data_from_selection_owner<'a>(
...
 			locked.shared.conn.as_ref().unwrap().flush();
```
ArturKovacs, line 802: And `log:error!` or `log::warn!` here as well.

src/x11_clipboard.rs
```diff
@@ -762,112 +643,83 @@ fn handle_selection_request_event(event: *mut xcb_selection_request_event_t) {
...
-	unsafe {
-		xcb_send_event(
-			shared.conn.as_ref().unwrap().get_raw_conn(),
-			0,
-			requestor,
-			XCB_EVENT_MASK_NO_EVENT,
-			&notify as *const _ as *const _,
-		)
-	};
+	shared.conn.as_ref().unwrap().send_event(false, requestor, EventMask::NO_EVENT, notify);
```
ArturKovacs, line 655: And `log::error!` or `log::warn!` here as well.

ArturKovacs (changes requested):

src/x11_clipboard.rs
daxpedda: This makes a lot of sense, I misunderstood what creating the window here does.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done. I just added `log::error!("{}", e)`, let me know if you want to add any more text to it.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done.

src/x11_clipboard.rs
daxpedda: Done.

ArturKovacs:
Awesome, thanks a lot! Oh and I almost forgot: yeah I also thought about significantly refactoring this, especially because how the locking and the condvar is used. It feels messy, but it would require me to gain a much better understanding of how this protocl works, and there are other things that keep me busy so I don't know if I'll ever be able to do this.

daxpedda:
No worries, I am in the same position.
I believe there is some potential to collaborate with @hecrj through [window_clipboard](https://crates.io/crates/window_clipboard), but he does seem quiet busy with other things too.

ArturKovacs:
That's a great idea! I would be up for some collaboration. There is a tough question though: which repo do we collaborate on? I would like to keep improving this, and I'm guessing @hecrj would like to stick to `window_clipboard`. Although it would probably be better for the ecosystem to reduce the amount of competing solutions for the exact same thing. I'm open to suggestions.

Side note: I'd be happy to add support for creating the clipboard from a window.

daxpedda:
I obviously can't speak for @hecrj or his project, but I know that the two projects divert in their scope.
But I agree, ideally we have only one implementation, preferably with the wider scope of course.

I think it's best if we wait for an answer from @hecrj or you contact him directly. I am also up for contributing and maintaining, as long as somebody helps with the X11 implementation :smile:.

hecrj:
Hey! I didn't know about this crate!

I started [`window_clipboard`] as an experiment and to satisfy any use cases that [`iced`] may have.

I am open to collaborate. This crate seems to be further ahead than [`window_clipboard`], so maybe I should simply ditch that and help here.

When it comes to [`iced`], I am concerned with the blocking time of the read operations. I started [`window_clipboard`] because the other available clipboard crates waited a long time while polling, causing dropped frames. Although I believe a proper solution would involve a proper concurrent architecture for the widget logic in [`iced`].

[`iced`]: https://github.com/hecrj/iced

daxpedda:
> Although I believe a proper solution would involve a proper concurrent architecture for the widget logic in [`iced`](https://github.com/hecrj/iced) (so they can read the clipboard without blocking everything else).

How would that work, do you mean an `async` interface?

ArturKovacs:
@hecrj I looked a little bit into your implementation and it seems to be much cleaner than the one here. There is something that I don't understand though. You have a separate window for reading and writing, and if I'm not mistaken you only take out events form your read event queue when a clipboard read happens. Is this okay to do? Shouldn't the event queue be polled constantly to avoid the queue growing too large?

My understanding on X programming is very limited so that's why I don't understand if/why this is okay.

PR #10, opened 2021-03-08
Remove prerequisites

daxpedda:
I believe this isn't needed anymore, I confirmed it in a virtual machine.


ArturKovacs:
How do you test this? (E.g. which distribution + desktop environment does your VM have?) For an other project of mine, I set up a virtual NixOS and made a nix expression to test which dependencies are sufficient, but that's a quite cumbersome approach for me considering that nix expressions are still mostly black magic to me.

ArturKovacs:
If you tried it on a clean installation and it worked without the dependency, then it's probably good. If you installed some devy stuff other than rust itself on these machines then I'm not sure how accurate this is.

ArturKovacs:
I think I'm tired. If you know for sure that this pacakge wasn't isntalled (simply by checking whether it was installed) at the time of compiling, then it's perfect.

daxpedda:
It was a fresh installation and I did check that the package wasn't installed.
I understand your tiredness :)

PR #12, opened 2021-04-03
Wayland clipboard support via data_control protocol

xanderio:
While it is true that wayland can support that X11 way of handling
clipboard support, this requires a running xwayland server and
compositor support for running xwayland.

This change introduces a new clipboard context
`WaylandDataControlClipboardContext` which is a wrapper around the
wl-clipboard-rs crate. This crate handles all of the communication with
the compositor for us.

To selected between X11 and wayland data_control clipboard handle is not
possible via a platform selector, as is the case for windows and mac,
two new features where add `x11` and `wayland-data-control`. The `x11`
feature is in the `default` feature set as to preserve backward
compatibility.

The feature name `wayland-data-control` was chosen to handle other
possible wayland protocols to interface with a clipboard.

Looking forward to your review :)

### DIFF
--- a/src/wayland_data_control_clipboard.rs
+++ b/src/wayland_data_control_clipboard.rs
@@ -1,10 +1,6 @@
-use super::common::{Error, ImageData};
+use super::common::{convert_to_png, Error, ImageData};
 
-use std::{
-	cell::RefCell,
-	io::{Cursor, Read},
-	rc::Rc,
-};
+use std::io::{Cursor, Read};
 
 use wl_clipboard_rs::copy::{Options, Source};
 use wl_clipboard_rs::paste::{get_contents, ClipboardType, Error as PasteError, Seat};
@@ -82,51 +78,15 @@ impl WaylandDataControlClipboardContext {
 	pub fn set_image(&mut self, image: ImageData) -> Result<(), Error> {
 		use wl_clipboard_rs::copy::MimeType;
 
-		/// This is a workaround for the PNGEncoder not having a `into_inner` like function
-		/// which would allow us to take back our Vec after the encoder finished encoding.
-		/// So instead we create this wrapper around an Rc Vec which implements `io::Write`
-		#[derive(Clone)]
-		struct RcBuffer {
-			inner: Rc<RefCell<Vec<u8>>>,
-		}
-		impl std::io::Write for RcBuffer {
-			fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-				self.inner.borrow_mut().extend_from_slice(buf);
-				Ok(buf.len())
-			}
-			fn flush(&mut self) -> std::io::Result<()> {
-				// Noop
-				Ok(())
-			}
-		}
-
-		if image.bytes.is_empty() || image.width == 0 || image.height == 0 {
-			return Err(Error::ConversionFailure);
-		}
-
-		let buffer = RcBuffer { inner: Rc::new(RefCell::new(Vec::new())) };
-		let encoding_result;
-		{
-			let encoder = image::png::PngEncoder::new(buffer.clone());
-			encoding_result = encoder.encode(
-				image.bytes.as_ref(),
-				image.width as u32,
-				image.height as u32,
-				image::ColorType::Rgba8,
-			);
-		}
-		// Rust impl: The encoder must be destroyed so that it lets go of its reference to the
-		// `output` before we `try_unwrap()`
-		if encoding_result.is_ok() {
+		convert_to_png(image).and_then(|image| {
 			let opts = Options::new();
-			let source = Source::Bytes(Rc::try_unwrap(buffer.inner).unwrap().into_inner().into());
+			let source = Source::Bytes(image.bytes.into());
 			opts.copy(source, MimeType::Specific("image/png".into())).map_err(into_unknown)?;
-		}
-
-		Ok(())
+			Ok(())
+		})
 	}
 }
 
-fn into_unknown(error: wl_clipboard_rs::copy::Error) -> Error {
+fn into_unknown<E: std::fmt::Display>(error: E) -> Error {
 	Error::Unknown { description: format!("{}", error) }
 }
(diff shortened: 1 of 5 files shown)


ArturKovacs:
Hey, thanks for the contribution!

Which compositor/desktop environment is this expected to work on? I tried with a KDE session and it failed

xanderio:
This protocol is from wlroots. So every compositor based on this will work e.g. sway, hikari.

https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-data-control-unstable-v1.xml

src/wayland_data_control_clipboard.rs
```diff
@@ -0,0 +1,132 @@
...
+
+			Err(err) => return Err(Error::Unknown { description: format!("{}", err) }),
+		}
+	}
+
+	pub fn set_text(&mut self, text: String) -> Result<(), Error> {
+		use wl_clipboard_rs::copy::MimeType;
+		let opts = Options::new();
+		let source = Source::Bytes(text.as_bytes().into());
+		opts.copy(source, MimeType::Autodetect).map_err(map_copy_error)?;
```
ArturKovacs, line 43: It seems to me that using `MimeType::Text` is more desirable here.

src/wayland_data_control_clipboard.rs
```diff
@@ -0,0 +1,132 @@
...
+		Ok(())
+	}
+
+	pub fn get_image(&mut self) -> Result<ImageData, Error> {
+		use wl_clipboard_rs::paste::MimeType;
+		let result = get_contents(ClipboardType::Regular, Seat::Unspecified, MimeType::Any);
```
ArturKovacs, line 49: I'm pretty sure this should use `MimeType::Specific("image/png")`.

src/wayland_data_control_clipboard.rs
```diff
@@ -0,0 +1,132 @@
...
+		impl std::io::Write for RcBuffer {
+			fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+				self.inner.borrow_mut().extend_from_slice(buf);
+				Ok(buf.len())
+			}
+			fn flush(&mut self) -> std::io::Result<()> {
+				// Noop
+				Ok(())
+			}
+		}
```
ArturKovacs, line 101: Considering that this is used from both x11 and wayland, and that this only depends on the standard library, I think this should be moved to `common.rs`.

src/wayland_data_control_clipboard.rs
```diff
@@ -0,0 +1,132 @@
...
+			let encoder = image::png::PngEncoder::new(buffer.clone());
+			encoding_result = encoder.encode(
+				image.bytes.as_ref(),
+				image.width as u32,
+				image.height as u32,
+				image::ColorType::Rgba8,
+			);
+		}
+		// Rust impl: The encoder must be destroyed so that it lets go of its reference to the
+		// `output` before we `try_unwrap()`
```
ArturKovacs, line 119: The "Rust impl" prefix should be removed from here. I only used it in the x11 implementation because many comments there are from the original C++ implementation.

ArturKovacs (changes requested):

src/wayland_data_control_clipboard.rs
```diff
@@ -0,0 +1,132 @@
...
+			let opts = Options::new();
+			let source = Source::Bytes(Rc::try_unwrap(buffer.inner).unwrap().into_inner().into());
+			opts.copy(source, MimeType::Autodetect).map_err(map_copy_error)?;
+		}
+
+		Ok(())
+	}
+}
+
+fn map_copy_error(error: wl_clipboard_rs::copy::Error) -> Error {
```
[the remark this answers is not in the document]
xanderio: I would love to do so, but `wl-clipboard-rs::copy::Error` dos not implement `std::error::Error`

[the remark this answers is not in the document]
ArturKovacs: Oh. And what about `Display` like this
```rust
fn into_unknown<E: std::fmt::Display>(error: E) -> Error {
    Error::Unknown {
        description: format!("{}", error),
    }
}
```

xanderio:
As i don't have a X11 setup, i wasn't able to verify the changes to x11_clipboard.rs

src/wayland_data_control_clipboard.rs
xanderio: Im not sure if we should just allow PNG image to be fetched from the clipboard. What about an JPEG image in the clipboard?

ArturKovacs: The problem here is that a clipboard may store a single "object" in multiple formats. For example if you copy some SVG elements using your SVG editor, it might place both `image/svg` and `image/png` onto the clipboard. The `get_contents` function only retrieves one of these items, so if we are unlucky we would get the SVG which we can't decode into an pixel array.

Considering that `image/png` is the de facto image format for the clipboard on Linux, we should ask specifically for `image/png`. (Also note that arboard only supports PNG currently. See the image dependency in `Cargo.toml`)

ArturKovacs (changes requested):
I also just realized that the way this is currently structured, there's no way to build `arboard` in a way that it supports *both* X11 and Wayland, but this should be allowed because most developers will want to make software be compatible with X11.

So there should be a struct which wraps around both the X11 and the Wayland clipboard and decides at runtime which one to use. This struct could be defined in the `linux_common.rs` module which I mention in one of my comments in this review.

ArturKovacs:
Considering that the changes I'm asking for are not trivial, and that I have a somewhat specific idea about how it should be done, if you'd like I can just merge this PR into a topic branch, and I can finish the rest of the work myself. But if you want to complete this, that's totally fine too.

xanderio:
Sure go ahead. If i can be of any other help to this progress please let me know. :)

ArturKovacs:
Awesome. I created an `add-wayland` branch. Just re-target this PR to that branch, and I'll take care of the rest. 🙂

ArturKovacs:
Thanks again!

PR #13, opened 2021-04-11
X11 rewrite

ArturKovacs:
A new X11 backend was written from scratch to improve performance and maintainability.

### DIFF
--- a/src/x11_clipboard.rs
+++ b/src/x11_clipboard.rs
@@ -22,7 +22,7 @@ use std::{
 	usize,
 };
 
-use log::{error, info, trace, warn};
+use log::{error, trace, warn};
 use once_cell::sync::Lazy;
 use parking_lot::{Condvar, Mutex, MutexGuard, RwLock};
 use x11rb::{
@@ -389,7 +389,7 @@ impl ClipboardContext {
 
 		// we found something
 		if reply.type_ == target_format {
-			return Ok(ReadSelNotifyResult::GotData(reply.value));
+			Ok(ReadSelNotifyResult::GotData(reply.value))
 		} else if reply.type_ == self.atoms.INCR {
 			// Note that we call the get_property again because we are
 			// indicating that we are ready to receive the data by deleting the
@@ -414,12 +414,12 @@ impl ClipboardContext {
 				let min_data_len = reply.value32().and_then(|mut vals| vals.next()).unwrap_or(0);
 				incr_data.reserve(min_data_len as usize);
 			}
-			return Ok(ReadSelNotifyResult::IncrStarted);
+			Ok(ReadSelNotifyResult::IncrStarted)
 		} else {
 			// this should never happen, we have sent a request only for supported types
-			return Err(Error::Unknown {
+			Err(Error::Unknown {
 				description: String::from("incorrect type received from clipboard"),
-			});
+			})
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


src/x11_clipboard.rs
```diff
@@ -8,1044 +8,858 @@ the Apache 2.0 or the MIT license at the licensee's choice. The terms
...
-				manager!().window = 0;
+				.map_err(into_unknown)?
+				.reply()
+				.map_err(into_unknown)?;
+			log::trace!("Receiving INCR segments");
+			*using_incr = true;
+			if reply.value_len == 4 {
+				let min_data_len = reply.value32().and_then(|mut vals| vals.next()).unwrap_or(0);
+				incr_data.reserve(min_data_len as usize);
 			}
```
daxpedda, line 438: What exactly does it mean if it is **not** a `u32`?

src/x11_clipboard.rs
```diff
@@ -8,1044 +8,858 @@ the Apache 2.0 or the MIT license at the licensee's choice. The terms
...
-		property,
-	};
-	if let Err(e) =
-		shared.conn.as_ref().unwrap().send_event(false, requestor, EventMask::NO_EVENT, notify)
-	{
-		log::error!("{}", e)
-	}
-	if let Err(e) = shared.conn.as_ref().unwrap().flush() {
-		log::error!("{}", e)
+		warn!("The handover was not finished and the condvar didn't time out, yet the condvar wait ended. This should be unreachable.");
```
daxpedda, line 622: If this is truly correct, we should use `unreachable!`?

daxpedda (changes requested):
I'm quiet frustrated after going through this. This is kind of terrible, but I can't see any other way around it.

Thank you for all the work actually completing our implementation. I think you achieved the best we could do here.

src/x11_clipboard.rs
ArturKovacs: It is truly correct if I understand correctly, although I don't think that panicking would be good here. Panicking is useful in situations where it's completely meaningless for the program to try to continue. However even if something went wrong here, we don't gain anything from panicking but other code which might be very important won't get to run.

daxpedda: I was assuming that if this point was reached, something really went wrong with our code. But you are right, in any case, it would be useful to return an `Err` here instead of continuing like nothing happened.

src/x11_clipboard.rs
ArturKovacs: ICCCM says (at https://tronche.com/gui/x/icccm/sec-2.html#s-2.7.2)
> The contents of the INCR property will be an integer, which represents a lower bound on the number of bytes of data in the selection.

Although it does not specify the size of that integer (because why would it 🙄). Considering that the `int` type in C is 32 bits wide on both the x86 and the x86_64 targets, I'm assuming that they mean a 32 bit integer.

However it's not completely impossible that another process sends a 16 bit integer or a 64bit integer (if that's even possible). In such cases all that happens is that we don't allocate the buffer up-front but everything will work just the same, we will just incrementally allocate the buffer as the chunks come in.

daxpedda: https://www.x.org/releases/current/doc/xorg-docs/icccm/icccm.html#Use_of_Selection_Properties
Says the format is 32-bit?

ArturKovacs: Nice find! Yeah this is reassuring. And still even if it's not 32bit sometimes, nothing would go wrong.

ArturKovacs:
> I'm quiet frustrated after going through this. This is kind of terrible, but I can't see any other way around it.

Yeah, no surprise so many are excited about Wayland. (Although I actually don't know how the clipboard is handled in Wayland.) It is annoying but this PR is a huge improvement on the previous implementation. The previous was shameful but this is something that I am proud of, given the weird design of X.

PR #18, opened 2021-07-06
Make image data functionality an optional feature

dnaka91:
Currently the `image` crate and platform specific additional crates & features are always compiled even if the image related feature is not used. In my current project I have exactly this case where I only need to set text for the clipboard and don't really care about images (although that feature is really neat!).

This PR adds a new `image-data` feature that will only enable the image related functionality when activated, thus greatly reducing the amount of dependencies and therefore build time when that feature isn't needed.

This would currently be a breaking API change as it's now disabled by default and image features would disappear for upgrading users until they add the feature flag. If you like these changes but would like to make it non-breaking, I can add the `image-data` feature to the `default` set of course 🙇‍♂️.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,17 +39,17 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --verbose
+          args: --verbose --no-default-features
       - name: Run `cargo clippy` with `image-data` feature
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --verbose --features image-data
+          args: --verbose --no-default-features --features image-data
       - name: Run `cargo clippy` with `wayland-data-control` feature
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --verbose --features wayland-data-control
+          args: --verbose --no-default-features --features wayland-data-control
       - name: Run `cargo clippy` with all features
         uses: actions-rs/cargo@v1
         with:
@@ -74,15 +74,19 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+          args: --no-default-features
       - name: Run tests with `image-data` feature
         uses: actions-rs/cargo@v1
         with:
-          command: test --features image-data
+          command: test
+          args: --no-default-features --features image-data
       - name: Run tests with `wayland-data-control` feature
         uses: actions-rs/cargo@v1
         with:
-          command: test --features wayland-data-control
+          command: test
+          args: --no-default-features --features wayland-data-control
       - name: Run tests with all features
         uses: actions-rs/cargo@v1
         with:
-          command: test --all-features
+          command: test
+          args: --all-features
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -10,6 +10,7 @@ keywords = ["clipboard", "image"]
 edition = "2018"
 
 [features]
+image-data = ["core-graphics", "image", "winapi/minwindef", "winapi/wingdi", "winapi/winnt"]
```
ArturKovacs, line 13: As you suggested, let's make `image-data` a default feature.

Cargo.toml
```diff
@@ -22,28 +23,33 @@ env_logger = "0.8.3"
...
 parking_lot = "0.11"
 once_cell = "1.7"
+
+[[example]]
+name = "get_image"
+required-features = ["image-data"]
+
+[[example]]
+name = "set_image"
+required-features = ["image-data"]
```
ArturKovacs, line 55: When `image-data` becomes a default feature, I believe these won't be needed.

.github/workflows/test.yml
```diff
@@ -24,15 +24,33 @@ jobs:
...
+      - name: Run `cargo clippy` with `image-data` feature
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --verbose --features image-data
+      - name: Run `cargo clippy` with `wayland-data-control` feature
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --verbose --features wayland-data-control
```
ArturKovacs, line 52: When `image-data` becomes a default, these should be adjusted so that they are called with `--no-default-features`

.github/workflows/test.yml
```diff
@@ -52,8 +70,19 @@ jobs:
...
+        with:
+          command: test --features image-data
+      - name: Run tests with `wayland-data-control` feature
+        uses: actions-rs/cargo@v1
+        with:
+          command: test --features wayland-data-control
+      - name: Run tests with all features
+        uses: actions-rs/cargo@v1
+        with:
+          command: test --all-features
```
ArturKovacs, line 88: Same here as above: `--no-default-features`

ArturKovacs (changes requested):
Thanks, great PR. The only thing really is just making the feature a default and the things that come with that.

.github/workflows/test.yml
dnaka91: There is a nice tool [cargo-hack](https://github.com/taiki-e/cargo-hack) for it that conveniently runs commands on all combinations of the crate and avoids missing out combinations in the future.

How do you feel about using it here? I use it myself quite often, just in the CI it might take a bit of extra time to compile it first.

Cargo.toml
dnaka91: These would still be needed as you can use `--no-default-features` and then they would fail to compile.

ArturKovacs: Ah, gotcha!

.github/workflows/test.yml
ArturKovacs: Let's just keep it manual for now.

ArturKovacs:
Thanks for the contribution!

PR #20, opened 2021-08-06
Update all non-dev dependencies

complexspaces:
Arboard used to depend on two yanked crate versions, both of which came through `clipboard-win`.

This PR bumps it, and `core-graphics`, to their latest version to remove all the yanked crates from `arboard`s default dependency tree.


ArturKovacs:
Thanks a lot for the contribution!

PR #22, opened 2021-09-09
Windows CF_DIBV5 implement

everplus:
Windows CF_DIBV5 implement


ArturKovacs:
Thanks a lot! It's always very nice to see someone contributing code to solve an issue they found.

Although I do have to look into this a bit more, because I would prefer to use the winapi structures and constants and I would like to make sure that both `get_image` and `set_image` is correct.

everplus:
Thanks for your reply. This is a great learning opportunity.

PR #23, opened 2021-09-13
Fix incorrect `get_image` and `set_image` implementation

### LINKED ISSUE #21: Pasted image in Word can’t be displayed，but mspaint can
Pasted image in Word can’t be displayed，but mspaint can

Steps to reproduce

1. Right click to copy any image in Chrome
2. Run the program
``` rust
extern crate arboard;

use arboard::{Clipboard, ImageData};
use std::error::Error;

fn main() -> Result<(), Box<dyn Error>> {
    let mut clipboard = Clipboard::new()?;
    let image: ImageData = clipboard.get_image()?.to_owned_img();
    println!(
        "Found image, width: {}px, height: {}px, size: {}bytes",
        image.width,
        image.height,
        image.bytes.len()
    );
    clipboard.set_image(image)?;
    return Ok(());
}
```
3. Paste in Word

ArturKovacs:
Fixes #21

### DIFF
--- a/src/windows_clipboard.rs
+++ b/src/windows_clipboard.rs
@@ -9,7 +9,6 @@ and conditions of the chosen license apply to this file.
 */
 
 #[cfg(feature = "image-data")]
-use std::io::{self, Read, Seek};
 use std::mem::size_of;
 
 use clipboard_win::Clipboard as SystemClipboard;
@@ -18,9 +17,6 @@ use scopeguard::defer;
 use winapi::um::{
 	stringapiset::WideCharToMultiByte,
 	winbase::{GlobalLock, GlobalSize, GlobalUnlock},
-	wingdi::{
-		GetDIBits, BITMAPINFO, BITMAPINFOHEADER, BI_RGB, PROFILE_EMBEDDED, PROFILE_LINKED, RGBQUAD,
-	},
 	winnls::CP_UTF8,
 	winuser::{GetClipboardData, CF_UNICODETEXT},
 };
@@ -29,8 +25,9 @@ use winapi::{
 	shared::minwindef::DWORD,
 	um::{
 		wingdi::{
-			CreateDIBitmap, DeleteObject, LCS_sRGB, BITMAPV4HEADER, BITMAPV5HEADER, BI_BITFIELDS,
-			CBM_INIT, CIEXYZTRIPLE, DIB_RGB_COLORS, LCS_GM_IMAGES,
+			CreateDIBitmap, DeleteObject, GetDIBits, LCS_sRGB, BITMAPINFO, BITMAPINFOHEADER,
+			BITMAPV4HEADER, BITMAPV5HEADER, BI_BITFIELDS, BI_RGB, CBM_INIT, CIEXYZTRIPLE,
+			DIB_RGB_COLORS, LCS_GM_IMAGES, PROFILE_EMBEDDED, PROFILE_LINKED, RGBQUAD,
 		},
 		winnt::LONG,
 		winuser::{GetDC, SetClipboardData, CF_BITMAP},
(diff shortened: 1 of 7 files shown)


ArturKovacs:
@everplus here's al alternative solution to the issue you encountered. Thanks again for your PR, it helped me testing things and getting to this solution.

Please test this and make sure it solves the problem for you.

everplus:
The dibv5 image was set incorrectly. I am not familiar with winapi, I use a third-party tool to find that all bytes are zero. 

image
image

Below is the result of my PR：

image

ArturKovacs:
Thank you for testing; I did not realize this was happening.

Another way to reproduce this issue is to run something like
```rust
    let mut clipboard = Clipboard::new()?;
    let image: ImageData = clipboard.get_image()?.to_owned_img();
    clipboard.set_image(image)?;

    // This fails
    clipboard.get_image()?;
```

I'm going to fix this later.

I will also change the declaration of `get_image` to return an `ImageData<'static>` because it's owned anyways, just the compiler doesn't know about it.

ArturKovacs:
I think I found what the issue was. Turns out that when a program sets the `CF_BITMAP` and another requests `CF_DIBV5`, then Windows "synthesizes" the `CF_DIBV5` from the `CF_BITMAP` (it also synthesizes `CF_BITMAP` from `CF_DIBV5`). But arboard tried to set both the DIBV5 and the BITMAP which apparently Windows didn't like too much so only setting DIBV5 is enough the allow any other program to get the data in DIBV5 and BITMAP.

ArturKovacs:
@everplus could you test this one more time to make sure it works as it should?

everplus:
I tried PR, but it was still unsuccessful.
I read the code, and I feel that the problem seems to be CreateDIBitmap, maybe the DIBv5 format should use CreateDIBSection

ArturKovacs:
Hmm you are right, the data was still inaccessible. I *think* I finally fixed it because now "InsideClipboard" shows the contents. I'm suspecting that the `GlobalUnlock` function didn't run and so other processes couldn't get access to the contents of the image.

Could you test this again?

everplus:
image

I guess the useless bytes are generated by CreateDIBitmap

ArturKovacs:
It's kinda funny how much trouble this little issue is causing.

First of all let me reflect to your speculations:
- What you highlighted in the image as "resolution" is the physical word resolution (ie pixels per inch) and it's unrelated to the number of pixels in the image.
- `CreateDIBitmap` is almost certainly not related to placing incorrect data onto the clipboard, because that function is only called from `get_image`

According to my investigation, the problem this time was that we had the pixels in top-to-bottom order (negative height in the header), but MS Word cannot read such images. For MS Word, the image must be specified in bottom-to-top order (positive height in the header).

As always, I'd like you to test this again. It should really work this time, but I don't want to jinx it.

everplus:
Hmm... it's zero again...

everplus:
image
image

ArturKovacs:
That's very strange. I tested again and for me both the DIBV5 and the bitmap contains the correct image.

Could you describe what you did to get this or could you share a minimal example code that produces all zeroes?

There is some logging in arboard, so I'd like to ask you to run your program with logging enabled (for example with [simple_logger](https://crates.io/crates/simple_logger)) and upload the logs here for when it produces zeroes.

everplus:
[dib_image_convert.zip](https://github.com/ArturKovacs/arboard/files/7233149/dib_image_convert.zip)

I uploaded my test code.
Steps to reproduce:
1. Open link in Chrome, `https://pic.c-ctrip.com/platform/online/login/c_logo2020.png`
2. Right click to copy the image
2. Run the program, `cargo run`
3. Paste in Microsoft Word (failed)
4. Paste in Mspaint (failed, just 1*1)
4. Check in FreeClipViewer (invalid, all zeroes)

ArturKovacs:
Thank you, using the instructions you provided, I could reproduce the behavior when only zeroes appear on the clipboard.

There is some serious black magic happening here, so after spending some time investigating, I decided to stop for now.

What I found is that if you just add `simple_logger = "1.11"` as a dependency to `Cargo.toml` and run again with `cargo run`, then it will work. Even though the produced exe file is identical (I checked with `comp`). Furthermore, using my own example, when I ran the exe directly (without cargo), it worked. Using your example code it didn't work when I ran the exe directly. This means that the problem can still come up when *not* running with cargo.

Honestly I'm not entirely sure how we could find the cause of this problem, but my best guess right now is to read the documentation about [using the windows clipboard](https://docs.microsoft.com/en-us/windows/win32/dataxchg/using-the-clipboard).

ArturKovacs:
I tried creating a hidden window just to own the clipboard, but that didn't help either.

src/windows_clipboard.rs
```diff
@@ -113,13 +113,13 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
...
 
-		let pixels_dst = data_ptr.add(header_size);
-		copy_nonoverlapping(image.bytes.as_ptr() as *const _, pixels_dst, image.bytes.len());
+		// Not using the `add` function, because that has a restriction, that the result cannot overflow isize
+		let pixels_dst = (data_ptr as usize + header_size) as *mut u8;
+		copy_nonoverlapping::<u8>(image.bytes.as_ptr(), pixels_dst, image.bytes.len());
 
-		let dst_pixels_slice =
-			std::slice::from_raw_parts_mut(pixels_dst as *mut u8, image.bytes.len());
+		let dst_pixels_slice = std::slice::from_raw_parts_mut(pixels_dst, image.bytes.len());
```
ArturKovacs, line 122: This was the solution to all that headache and mysterious anomalies.

ArturKovacs:
@everplus I know it's been a long time, but I'm confident that the issue is really fixed this time. Can I ask you to test this to make sure?

everplus:
The problem is solved. Thank you very much.

PR #29, opened 2021-11-03
Reassert X.Org clipboard ownership on data change

mathieulj:
# What

When we change the clipboard data repeatedly, re-assert ownership of the clipboard even if we are still the owner.

# Why

The [ICCCM suggests that client windows should re-assert ownership of the clipboard when data changes](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#responsibilities_of_the_selection_owner#the_clipboard_selection). Including the relevent part here to save you from having to look through that document (the emphasis is my own):
> Whenever a client wants to transfer data to the clipboard:
> - It should assert ownership of the CLIPBOARD.
> - If it succeeds in acquiring ownership, it should be prepared to respond to a request for the contents of the CLIPBOARD in the usual way (retaining the data to be able to return it). The request may be generated by the clipboard client described below.
> - If it fails to acquire ownership, a cutting client should not actually perform the cut or provide feedback that would suggest that it has actually transferred data to the clipboard.
>
> **The owner should repeat this process whenever the data to be transferred would change.**

# Symptom of not doing this

Some clients that need to keep up to date with the clipboard (ex. [TigerVNC](https://github.com/TigerVNC/tigervnc)) poll for the last timestamp of the atom to know if the contained data has changed. In absence of re-asserting our ownership, the last timestamp remains unchanged so those clients will not request the new data even if we have some changes available. One way to have this manifests itself is to open a TigerVNC session while you have the clipboard open on your local machine and periodically updating (i.e. hang onto the clipboard instance and periodically `.set_text(new_data)`), you could then paste into the VNC session but TigerVNC would only sync the first `set_text` after ownership changes (ex. every time you, Ctrl+C in another app, the next `set_text` is synced).

### DIFF
--- a/src/x11_clipboard.rs
+++ b/src/x11_clipboard.rs
@@ -166,7 +166,7 @@ impl XContext {
 	}
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 struct ClipboardData {
 	bytes: Vec<u8>,
 
@@ -205,8 +205,19 @@ impl ClipboardContext {
             });
 		}
 
-		// only if the current owner isn't us, we try to become it
-		if !self.is_owner(selection)? {
+		let data_changed = {
+			// Guard encapsulated in a block so that it is dropped before making the requests.
+			let mut guard = self.data_of(selection).write();
+
+			// Just setting the data, and the `serve_requests` will take care of the rest.
+			let previous_data = guard.replace(data);
+
+			previous_data == *guard
+		};
+
+		// ICCCM version 2, section 2.6.1.3 states that we should re-assert ownership whenever data
+		// changes.
+		if data_changed || !self.is_owner(selection)? {
 			let server_win = self.server.win_id;
 
 			self.server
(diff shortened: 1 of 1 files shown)


src/x11_clipboard.rs
```diff
@@ -205,8 +205,19 @@ impl ClipboardContext {
...
-		// only if the current owner isn't us, we try to become it
-		if !self.is_owner(selection)? {
+		let data_changed = {
+			// Guard encapsulated in a block so that it is dropped before making the requests.
+			let mut guard = self.data_of(selection).write();
+
+			// Just setting the data, and the `serve_requests` will take care of the rest.
+			let previous_data = guard.replace(data);
+
+			previous_data == *guard
```
ArturKovacs, line 215: Only re-asserting the ownership when the data is actually different is a good way to save some CPU cycles given that we work with small data. However if the clipboard stores bigger data, like images, then this is likely a waste of CPU cycles.

I think that unconditionally re-asserting ownership is best here.

ArturKovacs (changes requested):
Thanks a lot for the PR!

src/x11_clipboard.rs
mathieulj: I was not sure about guarding for a change at first. You are correct that it will cost even higher for large datasets and this code path is only hit when users of this crate  ask for a write so I think it's safe to remove and will do so.

ArturKovacs:
Big thanks!

PR #33, opened 2022-03-07
Dependency updates

complexspaces:
This updates most of `arboard`'s dependencies. I skipped updating `image` for now because it appears that `0.24` is having dependency conflicts with `criterion`.


Cargo.toml
```diff
@@ -18,8 +18,8 @@ wayland-data-control = ["wl-clipboard-rs"]
 thiserror = "1.0"
 
 [dev-dependencies]
-simple_logger = "1.11"
-env_logger = "0.8.3"
+simple_logger = "2.1"
```
kjvalencik, line 21: Should `Cargo.lock` be removed from `.gitignore` to catch `dev-dependencies`?

kjvalencik (approved):

PR #35, opened 2022-03-07
Windows clipboard cleanup + fixes

complexspaces:
This PR makes some slight cleanups to the Windows clipboard code, along with fixing an alignment soundness bug that Miri reported.

### Cleanup
The majority of the cleanup work was shifting much more responsibility to `win-clipboard` to handle the set/get operations. This allows `arboard` to drop a lot of its Windows-related unsafe blocks in favor of safe code. 

The final change was clarifying the safety of `add_cf_dibv5`. IMO, marking the entire function as `unsafe` was too broad and makes it harder to follow which operations inside are actually unsafe.

### Soundness Fix
While working on this, Miri reported 2 soundness problems.

You can reproduce the Miri finding by:
1. Check out the master branch
2. Copy the `check_win_to_rgba_conversion` test.
3. Run `cargo +nightly miri test check_win_to_rgba_conversion --features image-data`.

The error that I saw locally is the following:
```
cargo +nightly miri test check_win_to_rgba_conversion --features image-data

running 1 test
error: Undefined Behavior: accessing memory with alignment 1, but alignment 4 is required
   --> C:\Users\USER\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\slice\raw.rs:131:14
    |
131 |     unsafe { &mut *ptr::slice_from_raw_parts_mut(data, len) }
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ accessing memory with alignment 1, but alignment 4 is required
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information

    = note: inside `std::slice::from_raw_parts_mut::<u32>` at C:\Users\USER\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\slice\raw.rs:131:14
note: inside `windows_clipboard::win_to_rgba` at src\windows_clipboard.rs:282:18
   --> src\windows_clipboard.rs:282:18
    |
282 |     let u32pixels = std::slice::from_raw_parts_mut(bytes.as_mut_ptr() as *mut u32, bytes.len() / 4);
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `windows_clipboard::tests::check_win_to_rgba_conversion` at src\windows_clipboard.rs:445:12
   --> src\windows_clipboard.rs:445:12
    |
445 |         unsafe { win_to_rgba(&mut data) }
    |                  ^^^^^^^^^^^^^^^^^^^^^^
note: inside closure at src\windows_clipboard.rs:443:2
   --> src\windows_clipboard.rs:443:2
    |
442 |       #[test]
    |       ------- in this procedural macro expansion
443 | /     fn check_win_to_rgba_conversion() {
444 | |         let mut data = DATA;
445 | |         unsafe { win_to_rgba(&mut data) }
446 | |     }
    | |_____^
    = note: this error originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)

```

A similar error is reported in `rgba_to_win`, but only when running with `-Zmiri-symbolic-alignment-check` set. To fix both of the problems, I switched the code from constructing a new slice to using [`align_to_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.align_to_mut) to perform a more well-checked cast. If the safety conditions are upheld, then this is correct and won't cause any weird image bugs to occur.

While I was in the functions, I switched the pixel shifting to use safe code instead. They have [similar enough code](https://rust.godbolt.org/z/evWjrxsjr) that there isn't a performance impact from this change.

### DIFF
--- a/src/windows_clipboard.rs
+++ b/src/windows_clipboard.rs
@@ -39,7 +39,7 @@ use super::common::ImageData;
 const MAX_OPEN_ATTEMPTS: usize = 5;
 
 #[cfg(feature = "image-data")]
-unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
+fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 	use std::intrinsics::copy_nonoverlapping;
 	use winapi::um::{
 		winbase::{GlobalAlloc, GHND},
@@ -65,7 +65,8 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 		bV5BlueMask: 0x000000ff,
 		bV5AlphaMask: 0xff000000,
 		bV5CSType: LCS_sRGB as u32,
-		bV5Endpoints: std::mem::zeroed(),
+		// SAFETY: Windows ignores this field because `bV5CSType` is not set to `LCS_CALIBRATED_RGB`.
+		bV5Endpoints: unsafe { std::mem::zeroed() },
 		bV5GammaRed: 0,
 		bV5GammaGreen: 0,
 		bV5GammaBlue: 0,
@@ -82,7 +83,7 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 	let image = flip_v(image);
 
 	let data_size = header_size + image.bytes.len();
-	let hdata = GlobalAlloc(GHND, data_size);
+	let hdata = unsafe { GlobalAlloc(GHND, data_size) };
 	if hdata.is_null() {
 		return Err(Error::Unknown {
 			description: format!(
@@ -91,7 +92,7 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 			),
 		});
 	}
-	{
+	unsafe {
 		let data_ptr = GlobalLock(hdata) as *mut u8;
 		if data_ptr.is_null() {
 			return Err(Error::Unknown {
@@ -117,12 +118,18 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 		rgba_to_win(dst_pixels_slice);
 	}
 
-	if SetClipboardData(CF_DIBV5, hdata as _).is_null() {
-		DeleteObject(hdata as _);
-		return Err(Error::Unknown {
-			description: format!("Call to `SetClipboardData` returned NULL at line {}", line!()),
-		});
+	unsafe {
+		if SetClipboardData(CF_DIBV5, hdata as _).is_null() {
+			DeleteObject(hdata as _);
+			return Err(Error::Unknown {
+				description: format!(
+					"Call to `SetClipboardData` returned NULL at line {}",
+					line!()
+				),
+			});
+		}
 	}
+
 	Ok(())
 }
 
@@ -378,6 +385,6 @@ impl WindowsClipboardContext {
 			});
 		};
 
-		unsafe { add_cf_dibv5(image) }
+		add_cf_dibv5(image)
 	}
 }
--- a/src/windows_clipboard.rs
+++ b/src/windows_clipboard.rs
@@ -11,26 +11,20 @@ and conditions of the chosen license apply to this file.
 #[cfg(feature = "image-data")]
 use std::mem::size_of;
 
-use log::error;
-
 use clipboard_win::Clipboard as SystemClipboard;
 
+#[cfg(feature = "image-data")]
 use scopeguard::defer;
-use winapi::um::{
-	errhandlingapi::GetLastError,
-	stringapiset::WideCharToMultiByte,
-	winbase::{GlobalLock, GlobalSize, GlobalUnlock},
-	winnls::CP_UTF8,
-	winuser::{GetClipboardData, CF_UNICODETEXT},
-};
 #[cfg(feature = "image-data")]
 use winapi::{
 	shared::minwindef::DWORD,
 	um::{
+		errhandlingapi::GetLastError,
+		winbase::{GlobalLock, GlobalUnlock},
 		wingdi::{
 			CreateDIBitmap, DeleteObject, GetDIBits, LCS_sRGB, BITMAPINFO, BITMAPINFOHEADER,
-			BITMAPV5HEADER, BI_RGB, CBM_INIT, CIEXYZTRIPLE, DIB_RGB_COLORS, LCS_GM_IMAGES,
-			PROFILE_EMBEDDED, PROFILE_LINKED, RGBQUAD,
+			BITMAPV5HEADER, BI_RGB, CBM_INIT, DIB_RGB_COLORS, LCS_GM_IMAGES, PROFILE_EMBEDDED,
+			PROFILE_LINKED, RGBQUAD,
 		},
 		winnt::LONG,
 		winuser::{GetDC, SetClipboardData},
@@ -71,7 +65,7 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 		bV5BlueMask: 0x000000ff,
 		bV5AlphaMask: 0xff000000,
 		bV5CSType: LCS_sRGB as u32,
-		bV5Endpoints: std::mem::MaybeUninit::<CIEXYZTRIPLE>::zeroed().assume_init(),
+		bV5Endpoints: std::mem::zeroed(),
 		bV5GammaRed: 0,
 		bV5GammaGreen: 0,
 		bV5GammaBlue: 0,
@@ -109,7 +103,7 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 			if retval == 0 {
 				let lasterr = GetLastError();
 				if lasterr != 0 {
-					error!("Failed calling GlobalUnlock when writing dibv5 data. Error code was 0x{:X}", lasterr);
+					log::error!("Failed calling GlobalUnlock when writing dibv5 data. Error code was 0x{:X}", lasterr);
 				}
 			}
 		});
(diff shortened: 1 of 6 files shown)


src/windows_clipboard.rs
```diff
@@ -231,17 +234,22 @@ fn read_cf_dibv5(dibv5: &[u8]) -> Result<ImageData<'static>, Error> {
...
 #[allow(clippy::identity_op, clippy::erasing_op)]
 unsafe fn rgba_to_win(bytes: &mut [u8]) {
-	let u32pixels = std::slice::from_raw_parts_mut(bytes.as_mut_ptr() as *mut u32, bytes.len() / 4);
+	// Check safety invariants to catch obvious bugs.
+	debug_assert_eq!(bytes.len() % 4, 0);
+
+	// When the correct conditions are upheld, `std` returns everything in the well-aligned slice.
+	// Miri, however, will return all or part of it in the prefix/suffix when running with `-Zmiri-symbolic-alignment-check`.
+	// Miri is only used to check that our code is sound, so this is fine.
+	let (_, u32pixels, _) = bytes.align_to_mut::<u32>();
```
ArturKovacs, line 243: I didn't know this was a thing. This is awesome!

How do you know that this will never return a prefix/suffix when compiled to binary?

src/windows_clipboard.rs
```diff
@@ -279,142 +287,57 @@ fn flip_v(image: ImageData) -> ImageData<'static> {
...
 	pub(crate) fn get_text(&mut self) -> Result<String, Error> {
-		// Using this nifty RAII object to open and close the clipboard.
 		let _cb = SystemClipboard::new_attempts(MAX_OPEN_ATTEMPTS)
 			.map_err(|_| Error::ClipboardOccupied)?;
-		let mut result = String::new();
-		get_string(unsafe { result.as_mut_vec() })?;
-		Ok(result)
+
+		clipboard_win::get(clipboard_win::Unicode)
+			.map_err(|err| handle_op_error(err, "failed to read clipboard string"))
```
ArturKovacs, line 318: After this change, the `ContentNotAvailable` and `ConversionFailure` errors won't be reported from this function on Windows. I think that's not good.

src/windows_clipboard.rs
kjvalencik: The docs indicate that no such guarantee is made.

> The method may make the middle slice the greatest length possible for a given type and input slice, but only your algorithm’s performance should depend on that, not its correctness. It is permissible for all of the input data to be returned as the prefix or suffix slice.

src/windows_clipboard.rs
```diff
@@ -231,17 +234,22 @@ fn read_cf_dibv5(dibv5: &[u8]) -> Result<ImageData<'static>, Error> {
...
 	for p in u32pixels {
-		let tmp = *p;
-		let rgba = std::slice::from_raw_parts((&tmp) as *const u32 as *const u8, 4);
-		let a = (rgba[3] as u32) << (3 * 8);
-		let r = (rgba[0] as u32) << (2 * 8);
-		let g = (rgba[1] as u32) << (1 * 8);
-		let b = (rgba[2] as u32) << (0 * 8);
-
-		*p = a | r | g | b;
+		let [mut r, mut g, mut b, mut a] = p.to_ne_bytes().map(u32::from);
```
kjvalencik, line 246: I'm not clear how RGBA values are stored in memory as a u32. Are they in native endianness or should this be `to_le_bytes`?

src/windows_clipboard.rs
complexspaces: The `ContentNotAvailable` error should still be returned when applicable (and is on my system). You're correct that `ConversionFailure` can't be returned from this anymore but I wasn't able to immediately think of a situation where it would fail to convert UTF-8 to UTF-16.

src/windows_clipboard.rs
```diff
@@ -428,6 +351,34 @@ impl WindowsClipboardContext {
...
+	}
+}
+
+fn handle_op_error(err: clipboard_win::SystemError, msg: &'static str) -> Error {
+	if err.is_zero() {
+		Error::ContentNotAvailable
+	} else {
+		Error::Unknown { description: msg.into() }
+	}
+}
```
ArturKovacs, line 364: The meaning of this error code depends on the function it was returned from. Therefore it's incorrect to assume that an error code of zero always means that the content is not available.

src/windows_clipboard.rs
ArturKovacs: > The `ContentNotAvailable` error should still be returned when applicable (and is on my system)

See: https://github.com/1Password/arboard/pull/35#discussion_r822056589

> You're correct that `ConversionFailure` can't be returned from this anymore but I wasn't able to immediately think of a situation where it would fail to convert UTF-8 to UTF-16.

A Windows wide character string may contain invalid UTF16 (e.g. a single high surrogate without a low surrogate) so actually the original code wasn't perfect either I should have used the [`from_wide`](https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStringExt.html#tymethod.from_wide) method

complexspaces: Those are both good points, thanks! I'll take another look at this and see if it can be made more robust.

src/windows_clipboard.rs
```diff
@@ -71,7 +68,7 @@ unsafe fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 		bV5BlueMask: 0x000000ff,
 		bV5AlphaMask: 0xff000000,
 		bV5CSType: LCS_sRGB as u32,
-		bV5Endpoints: std::mem::MaybeUninit::<CIEXYZTRIPLE>::zeroed().assume_init(),
+		bV5Endpoints: unsafe { std::mem::zeroed() },
```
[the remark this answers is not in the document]
complexspaces: I've added a safety comment here, thanks for the idea.

src/windows_clipboard.rs
complexspaces: This method has been removed since it wasn't correct.

src/windows_clipboard.rs
complexspaces: @kjvalencik is correct that there isn't a guarantee made here. That being said, I don't know another alternative that doesn't involve lots of allocation on stable. `align_to_mut` seems to be the only method to perform this "safely", as transmuting the slice is returns a worse error from Miri about a dangling reference.

src/windows_clipboard.rs
complexspaces: @ArturKovacs Please take another look when you get a chance. I've changed this up to both properly detect the wrong content type and convert the string correctly.

src/windows_clipboard.rs
ArturKovacs: But it's not useful to be fast if the result is wrong. I suggest checking whether the size of the prefix or suffix is greater than zero, and if so, then allocate a new `Vec<u16>` for all the bytes instead, and use that instead of the return value of `align_to_mut`.

The signature of the function would change because we are potentially returning a newly allocated container
```rs
unsafe fn rgba_to_win(bytes: mut Vec<u8>) -> Vec<u8> {
```

src/windows_clipboard.rs
```diff
@@ -428,6 +396,26 @@ impl WindowsClipboardContext {
...
+	fn check_win_to_rgba_conversion() {
+		let mut data = DATA;
+		unsafe { win_to_rgba(&mut data) }
+	}
+
+	#[test]
+	fn check_rgba_to_win_conversion() {
+		let mut data = DATA;
+		unsafe { rgba_to_win(&mut data) }
 	}
```
ArturKovacs, line 420: I recommend avoiding to cast a `const` to `mut` in tests. This seems like an unnecessary safety hazard, especially because tests run in parallel by default.

ArturKovacs: Sorry I misread the code, this is completely fine, because you copy the data first. My bad.

src/windows_clipboard.rs
```diff
@@ -18,19 +18,16 @@ use clipboard_win::Clipboard as SystemClipboard;
 use scopeguard::defer;
 use winapi::um::{
 	errhandlingapi::GetLastError,
-	stringapiset::WideCharToMultiByte,
-	winbase::{GlobalLock, GlobalSize, GlobalUnlock},
-	winnls::CP_UTF8,
-	winuser::{GetClipboardData, CF_UNICODETEXT},
+	winbase::{GlobalLock, GlobalUnlock},
```
ArturKovacs, line 21: Note that in the File chaned tab, you can see a few warnings reported by the CI.

I think we can fix these warning by only importing these when the image-data feature is used.

src/windows_clipboard.rs
kjvalencik: It makes it more complicated, but returning a `Cow<u8>` would allow only allocating the `Vec` when necessary.

src/windows_clipboard.rs
complexspaces: These should be cleaned up now. In the process, I also found that the direct dependency on `scopeguard` was not needed since the Linux code already had a minimal run-on-drop structure available, which was reused.

src/windows_clipboard.rs
complexspaces: No worries, thanks for checking either way.

src/windows_clipboard.rs
complexspaces: I've now implemented a 100% safe fallback for the cases that we fail to get a perfectly aligned slice. It borrows some of @kjvalencik's thinking about utilizing `Cow` to only allocate in this worst case. The ideal path is still allocation-free for the processing.

src/windows_clipboard.rs
ArturKovacs: I looked at the code and it looks good to me now. Thanks for your work!

ArturKovacs (approved):

kjvalencik (approved):

PR #42, opened 2022-04-08
Remove `once_cell` dependency by using `const_mutex`

SabrinaJewson:
`parking_lot` has the nice property that mutexes can be constructed in a `const` context, allowing `Lazy` to be avoided here.


complexspaces (approved):
Thanks!

Issue #39, opened 2022-03-20
The copied data isn't available after the program exits on X11


ArturKovacs:
As was described at https://github.com/1Password/arboard/issues/37#issuecomment-1072407160

ArturKovacs:
@kjvalencik could you paste the example code and describe how to reproduce the issue?

guihkx:
Wait, isn't this by design on X11, though? 🤔 

I remember having to deal with this issue when I started Linux, and the response to this was always: "it's by design, use a clipboard manager".

To be fair, this was more than 8 years ago, so things could have changed now...

Also, this doesn't seem to affect all desktop environments. For instance, on GNOME, it appears that they have implemented a clipboard manager directly in their compositor (https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/320), which I assume, is there to keep the clipboard content persistent, like on Windows.

So, running @kjvalencik's  last example, the clipboard contents are actually preserved on my machine (by the way, there's a `;` at the end of line 3).

- Arch Linux
- GNOME 41.5
- X11 session

kjvalencik:
@guihkx That's interesting! I'm on a somewhat older version, Gnome 3.36.8; however, that merge request was added in 3.34 and should be present on my system. Installing gpaste did not seem to fix it.

So maybe it's not a bug in `arboard` and just a limitation of X11 and other software versions.

guihkx:
Hmm... Aren't you using Linux Mint, though?

As far as I'm aware, Mint only offers 3 desktop environments: Cinnamon, MATE and XFCE. They all have different window managers/compositors, and none of them uses mutter, from GNOME.

* Cinnamon: [Muffin](https://github.com/linuxmint/muffin)
* MATE: [Marco](https://github.com/mate-desktop/marco)
* XFCE: [xfwm4](https://gitlab.xfce.org/xfce/xfwm4)

PR #43, opened 2022-04-08
Support waiting on the clipboard to be overwritten on Linux

SabrinaJewson:
Note: Untested on Wayland.

Fixes: #39

### DIFF
--- a/examples/daemonize.rs
+++ b/examples/daemonize.rs
@@ -7,9 +7,13 @@ use arboard::ClipboardExtLinux;
 use simple_logger::SimpleLogger;
 use std::{env, error::Error, process};
 
+// An argument that can be passed into the program to signal that it should daemonize itself. This
+// can be anything as long as it is unlikely to be passed in by the user by mistake.
+const DAEMONIZE_ARG: &str = "__internal_daemonize";
+
 fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
 	#[cfg(target_os = "linux")]
-	if env::args().nth(1).as_deref() == Some("__internal_daemonize") {
+	if env::args().nth(1).as_deref() == Some(DAEMONIZE_ARG) {
 		Clipboard::new()?.set_text_wait("Hello, world!".into())?;
 		return Ok(());
 	}
@@ -18,7 +22,7 @@ fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
 
 	if cfg!(target_os = "linux") {
 		process::Command::new(env::current_exe()?)
-			.arg("__internal_daemonize")
+			.arg(DAEMONIZE_ARG)
 			.stdin(process::Stdio::null())
 			.stdout(process::Stdio::null())
 			.stderr(process::Stdio::null())
--- a/examples/daemonize.rs
+++ b/examples/daemonize.rs
@@ -18,7 +18,7 @@ fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
 		.stdin(process::Stdio::null())
 		.stdout(process::Stdio::null())
 		.stderr(process::Stdio::null())
-        .current_dir("/")
+		.current_dir("/")
 		.spawn()?;
 
 	Ok(())
--- a/examples/daemonize.rs
+++ b/examples/daemonize.rs
@@ -1,0 +1,25 @@
+//! Example showcasing the use of `set_text_wait` and spawning a daemon to allow the clipboard's
+//! contents to live longer than the process on Linux.
+
+use arboard::{Clipboard, ClipboardExtLinux};
+use simple_logger::SimpleLogger;
+use std::{env, error::Error, process};
+
+fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
+	if env::args().nth(1).as_deref() == Some("__internal_daemonize") {
+		Clipboard::new()?.set_text_wait("Hello, world!".into()).unwrap();
+		return Ok(());
+	}
+
+	let _logger = SimpleLogger::new().init().unwrap();
+
+	process::Command::new(dbg!(env::current_exe()?))
+		.arg("__internal_daemonize")
+		.stdin(process::Stdio::null())
+		.stdout(process::Stdio::null())
+		.stderr(process::Stdio::null())
+        .current_dir("/")
+		.spawn()?;
+
+	Ok(())
+}
--- a/examples/daemonize.rs
+++ b/examples/daemonize.rs
@@ -13,7 +13,7 @@ fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
 
 	let _logger = SimpleLogger::new().init().unwrap();
 
-	process::Command::new(dbg!(env::current_exe()?))
+	process::Command::new(env::current_exe()?)
 		.arg("__internal_daemonize")
 		.stdin(process::Stdio::null())
 		.stdout(process::Stdio::null())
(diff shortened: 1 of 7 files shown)


MattX:
Thanks, I think this would be really useful–under most X11 desktop environments, this is probably what most users want to call. For what it's worth, the code looks good to me!

Have you considered adding a timeout (probably fairly easy using `wait_timeout`)? I have a use case that would benefit from that: I want to give a few hundred ms to a clipboard manager to take the clipboard over, and just fail otherwise instead of hanging.

SabrinaJewson:
I did originally have `wait_timeout`, but removed it when I realized that `wl-clipboard-rs` doesn't support the operation. Maybe it's something that could be added in future though, if `wl-clipboard-rs` gets that functionality.

MattX:
Ah, that's unfortunate but it makes sense.

complexspaces:
Hey, thanks for the PR! I'll try and make some time this week to review what's here.

complexspaces (commented):
Overall, I think this seems like a solid proposal. I've left a few very small nits. I apologize for the delay getting around to this as well 😞.

 I'm curious to see if @arturmartins or @mathieulj have any strong opinions one way or another.

examples/daemonize.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use arboard::Clipboard;
+#[cfg(target_os = "linux")]
+use arboard::ClipboardExtLinux;
+use simple_logger::SimpleLogger;
+use std::{env, error::Error, process};
+
+fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
+	#[cfg(target_os = "linux")]
+	if env::args().nth(1).as_deref() == Some("__internal_daemonize") {
```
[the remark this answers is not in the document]
SabrinaJewson: Good idea!

SabrinaJewson:
I think I still prefer my version of that API, because to me it makes sense that setting the clipboard and waiting for the serving to complete should be a single atomic operation. The combined API also might enable a future optimization of not spawning another thread when it’s not necessary, which may be desirable.

Another option is to create a builder API [like `wl-clipboard-rs` has](https://docs.rs/wl-clipboard-rs/latest/wl_clipboard_rs/copy/struct.Options.html). This avoids the method duplication problem and the current functions can still coexist as convenience functions.

> With Wayland, we could just re-implement the thread spawn that it does internally based on if `foreground` is set or not.

What do you mean by this? Are you saying that we’d always set `foreground` to run it in our own background thread and then somehow communicate with that thread inside `wait_for_change`?

SabrinaJewson:
I’ve now added a builder API to help address the combinatorial explosion problem. Basically, instead of `clipboard.set_wait(text)` you will now call `clipboard.set().wait(text)`. A sketch of the new API is below:

```rs
impl Clipboard {
    // Same methods as before, now rendered shortcut methods
    fn get_text(&mut self) -> Result<String, Error> { self.get().text() }
    fn set_text(&mut self, text: String) -> Result<(), Error> { self.set().text(text) }
    fn get_image(&mut self) -> Result<ImageData<'static>, Error> { self.get().image() }
    fn set_image(&mut self, image: ImageData<'_>) -> Result<(), Error> { self.set().image(image) }

    fn get(&mut self) -> Get<'_>;
    fn set(&mut self) -> Set<'_>;
}

impl<'clipboard> Get<'clipboard> {
    fn text(self) -> Result<String, Error>;
    fn image(self) -> Result<ImageData<'static>, Error>;
    // Linux-only via GetExtLinux
    fn text_with_clipboard(self, selection: LinuxClipboardKind) -> Result<String, Error>;
}

impl<'clipboard> Set<'clipboard> {
    fn text(self, text: String) -> Result<(), Error>;
    fn image(self, image: ImageData<'_>) -> Result<(), Error>;
    // Linux-only via SetExtLinux
    fn wait(self) -> Self;
    fn text_with_clipboard(self, text: String, selection: LinuxClipboardKind) -> Result<(), Error>;
}
```

Also, I noticed that `x11_clipboard`, `wayland_data_control_clipboard`, `osx_clipboard` and `windows_clipboard` were declared `pub mod`s in the source code — I’m assuming this is a mistake, since they’re very low-level. This commit also makes them private. I break the API of `X11ClipboardContext` in `x11_clipboard` anyway so I think it’s for the better.

src/common_linux.rs
```diff
@@ -116,38 +120,28 @@ pub trait ClipboardExtLinux {
 }
 
 impl ClipboardExtLinux for super::Clipboard {
```
complexspaces, line 122: If a new major version is required anyway, I don't think there is much reason to keep this trait after the other new API additions. A user would already need to import a trait to use different clipboards so this one has no extra value AFAICT. Do you agree?

src/lib.rs
```diff
@@ -107,6 +173,15 @@ impl Clipboard {
 #[cfg(test)]
 #[test]
 fn all_tests() {
+	use std::{
+		sync::{
+			atomic::{self, AtomicBool},
+			Arc,
```
complexspaces, line 179: I believe this `Arc` import needs moved into the Linux specific tests block to clean resolve these warnings. Thanks for tidying these up too!

complexspaces (changes requested):
This looks great overall! Thank you for the extra time needed for implementing the builder pattern. 

I really like it and I think it will be useful for adding clipboard parameters to macOS and Windows in the future too.

complexspaces:
> Also, I noticed that x11_clipboard, wayland_data_control_clipboard, osx_clipboard and windows_clipboard were declared pub mods in the source code  — I’m assuming this is a mistake, since they’re very low-level...

That would be my best assumption, too. They seem like a remnant of the pattern that `rust-clipboard` originally had, but that would be a question for @ArturKovacs.

### Compatibility report:
- X11: None of the extra returned structures are useful at all. There are no APIs which use them. `X11ClipboardContext` is actually less-useful than the high-level `Clipboard` due to the lack of the extension trait.
- Wayland: `WaylandDataControlClipboardContext` can be constructed and used, but it doesn't provide any more features than the high-level `Clipboard`.
- macOS: `OSXClipboardCOntext` can't be constructed or used. There is also the `class` function, which removing from the public API would be a win too.
- Windows: `WindowsClipboardContext` can be constructed, but there is nothing a user could do with it.

complexspaces (approved):

src/common_linux.rs
```diff
@@ -166,56 +111,142 @@ impl LinuxClipboard {
...
 
-	/// Fetches utf-8 text from the clipboard and returns it.
-	pub fn get_text(&mut self) -> Result<String, Error> {
-		match self {
-			Self::X11(cb) => cb.get_text(),
+	pub(crate) fn text(self) -> Result<String, Error> {
+		self.text_with_clipboard(LinuxClipboardKind::Clipboard)
+	}
 
+	fn text_with_clipboard(self, selection: LinuxClipboardKind) -> Result<String, Error> {
```
mathieulj, line 129: This might not be usable for image data and departs a little further from the previous API but I'm interested in hearing your thoughts; could we also use the builder pattern for this one?

Ex. 
```rust	
pub(crate) fn new(clipboard: &'clipboard mut Clipboard) -> Self {
    Self { 
        clipboard,
        selection: LinuxClipboardKind::Clipboard
    }
}
```
```rust
fn clipboard(self, selection: LinuxClipboardKind) -> Self {
    Self {
        selection, 
        ..self
    }
}
```

```rust
ctx.set().clipboard(LinuxClipboardKind::Clipboard).text(TEXT1.to_string()).unwrap();
```

mathieulj (approved):
Thank you for being patient with me, great work.

src/common_linux.rs
SabrinaJewson: Originally I did have that API, but then I thought that because previously that because this library didn’t provide the functionality, X11 and Wayland didn’t support image data on primary and secondary clipboards. But it seems they do, so I’ve implemented that now!

PR #44, opened 2022-04-15
Make never-mutated param non-mut

MattX:
Pass the `using_incr` parameter to `handle_read_property_notify` by value instead of by mutable reference: it's never mutated and, being a `bool`, the pointer is longer than the data.

Also fix various typos in the code.


complexspaces (approved):

complexspaces:
Thanks for the cleanup work!

PR #50, opened 2022-05-26
Adds a dummy clipboard to fix building for unsupported platforms.

Yeicor:


complexspaces:
Hey there, thanks for the PR.

Would you be able to explain your use case for this? The intention currently is that users of `arboard` handle these platform availability differences in their own code.

It feels like if we merged this, it could very well lead to more confusion for users (or those looking) because the crate would compile for them, but only result in errors. This could also lead to annoyance. In my opinion, if an entire crate can't provide any functionality for a target, it probably should not compile for it.

Yeicor:
My issue was that my project had a transitive dependency that optionally enables the `arboard` feature, but it does so by default and I have no way to change it (without also patching it). I wanted to build my project for android but this `arboard` dependency, which I do not use at all, was one of the errors.

The fork that I made was enough to build my project for android, using the patch section on `Cargo.toml`. Hence, I have no need for this to actually be merged, but I thought that other users in the same situation may want this. I also get why the fail approach of not compiling for an unsupported platform may avoid confusion and how, apart from transitive dependencies, this may be fixed from the user code. Feel free to close the PR if you think it doesn't fit the project.

PR #53, opened 2022-06-09
Add public API futureproofing

complexspaces:
These changes will help make it easier in the future to modify parts of the API without requiring another major version increase.


src/common_linux.rs
```diff
@@ -66,6 +66,7 @@ pub fn encode_as_png(image: &ImageData) -> Result<Vec<u8>, Error> {
 /// See <https://specifications.freedesktop.org/clipboards-spec/clipboards-0.1.txt> for a better
 /// description of the different clipboards.
 #[derive(Copy, Clone, Debug)]
+#[non_exhaustive]
```
complexspaces, line 69: I am on the fence about this change, mostly because some users may find it useful. These haven't changed in the history of X11 and are essentially set in stone.

ArturKovacs: In my opinion `non_exhaustive` is very rarely warranted.

It forces the programmer to add a match arm that catches all unknown enum variants. But having a catch-all arm, prevents an error message showing up for when a new enum variant is added. And so the programmer will **not** get notified about a potential mistake/bug in the code.

I think that preventing bugs is more important than keeping the version number low.

For this reason, I think that `non_exhaustive` shouldn't be added here.

src/common.rs
```diff
@@ -159,3 +160,11 @@ impl<F: FnOnce()> Drop for ScopeGuard<F> {
 		}
 	}
 }
+
+/// Common trait for sealing platform extension traits.
+pub(crate) mod sealed {
```
kjvalencik, line 165: Slight favor of naming this `private` since it reads nice in the docs `private::Sealed` and matches the example given in the API guidelines.

https://rust-lang.github.io/api-guidelines/future-proofing.html

src/common_linux.rs
mathieulj: For this specific enum, I don't feel that it makes much of a difference either way. We only take this as an argument and never return it so odd are no external code will be matching on this. My first thought was slightly on the same side as @ArturKovacs for the one added to the error enum but, given that we already have a catchall like error variant there `Unknown`, I don't think there is a significant difference to end users. They already can't rely on exhaustiveness because they must have a generic branch that we could theoretically be stuffing any new error into 🤷.

complexspaces: To be honest, I could go either way on this myself. I don't mind at all removing `non_exhaustive` from the clipboard type. With the `Error` type, it feels a bit off to not have it. Users of 2.x.x and 1.x.x are used to needing, as @mathieulj called it, a catchall branch via `Error::Unknown`.

Regarding what you said @ArturKovacs, I believe that allowing more errors to be added with less versioning friction could lead to better downstream error handling over time. To give an example: If we consider the error today, we could change almost anything in the internals and simply bubble new or modified errors up as `Error::Unknown`. Users won't see that in the public API, and therefore won't become aware of it. Opposite of that, if we had the ability to add a variant it would show up on `docs.rs` and in various IDE's completions, giving it a greater chance to be noticed and the programmer to realize there's some new edge case that may affect them. I believe that would be much harder to accomplish with `String` based errors.

ArturKovacs: @mathieulj The Unknown variant should not be used to report well-defined error cases. It's only meant to provide debugging information and it's up-to change even between patch versions.

With that said, I think that adding `non_exhaustive` to `Error` is okay. I'm saying I don't think it should be added to `LinuxClipboardKind`.

src/common.rs
complexspaces: I have switched the name to `private` since there isn't a good reason not to follow the "reference" here, thanks for the feedback.

src/common_linux.rs
complexspaces: I've removed it from `LinuxClipboardKind` since I think we're all in agreement on that point. I appreciate the clarification on what you see the `Unknown` variant for as well since I think they'll be helpful in the future.

PR #55, opened 2022-06-09
Organize platform-specific code in `platform/`

SabrinaJewson:
This makes all the code for a specific platform easier to find because it’s all in one place. Moving the X11 and Wayland code into a Linux module additionally simplifies lots of the `#[cfg]`s because all the ones that say “enable this on Linux” are implied by the containing `linux` module rather than having to be restated on both the X11 and Wayland modules.

I also renamed `X11ClipboardContext` → `x11::Clipboard` and similarly `WaylandDataControlClipboardContext` → `wayland::Clipboard` to give each of those modules are more consistent API with every other platform-specific module.

### DIFF
--- a/src/platform/linux/x11.rs
+++ b/src/platform/linux/x11.rs
@@ -231,24 +231,24 @@ impl Inner {
 		self.server.conn.flush().map_err(into_unknown)?;
 
 		// Just setting the data, and the `serve_requests` will take care of the rest.
-		let clipboard = self.clipboard_of(selection);
-		let mut data_guard = clipboard.data.write();
+		let selection = self.selection_of(selection);
+		let mut data_guard = selection.data.write();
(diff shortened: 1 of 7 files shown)


src/platform/linux/x11.rs
```diff
@@ -382,7 +385,7 @@ impl ClipboardContext {
 		}
 	}
 
-	fn clipboard_of(&self, selection: LinuxClipboardKind) -> &Clipboard {
+	fn clipboard_of(&self, selection: LinuxClipboardKind) -> &Selection {
```
complexspaces, line 388: Nit: Should this method be renamed as well to match?

SabrinaJewson: Good point!

complexspaces (approved):

PR #57, opened 2022-06-22
Refactor and cleanup macOS implementation

complexspaces:
This PR cleans up and significantly refactors the macOS clipboard implementation to be be more idiomatic and easier to maintain.

Some notable changes include:
- Caching Objective-C classes to avoid the lookup cost on every operation.
- Refactoring array accessing
- Refactoring parameter assignment
- Improving result handling

### DIFF
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -18,18 +18,22 @@ use core_graphics::{
 	data_provider::{CGDataProvider, CustomData},
 	image::CGImage,
 };
-use objc::runtime::{Class, Object};
-#[cfg(feature = "image-data")]
-use objc::runtime::{BOOL, NO};
-use objc::{msg_send, sel, sel_impl};
-use objc_foundation::{INSArray, INSObject, INSString};
-use objc_foundation::{NSArray, NSDictionary, NSObject, NSString};
+use objc::{
+	msg_send,
+	runtime::{Class, Object},
+	sel, sel_impl,
+};
+use objc_foundation::{INSArray, INSObject, INSString, NSArray, NSDictionary, NSObject, NSString};
 use objc_id::{Id, Owned};
-use std::mem::transmute;
+use once_cell::sync::Lazy;
 
-// required to bring NSPasteboard into the path of the class-resolver
+// Required to bring NSPasteboard into the path of the class-resolver
 #[link(name = "AppKit", kind = "framework")]
 extern "C" {}
+
+static NSSTRING_CLASS: Lazy<&Class> = Lazy::new(|| Class::get("NSString").unwrap());
+#[cfg(feature = "image-data")]
+static NSIMAGE_CLASS: Lazy<&Class> = Lazy::new(|| Class::get("NSImage").unwrap());
 
 /// Returns an NSImage object on success.
 #[cfg(feature = "image-data")]
@@ -45,7 +49,7 @@ fn image_from_pixels(
 		height: CGFloat,
 	}
 
-	#[derive(Debug, Clone)]
+	#[derive(Debug)]
 	struct PixelArray {
 		data: Vec<u8>,
 	}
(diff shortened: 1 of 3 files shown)


src/platform/osx.rs
```diff
@@ -269,24 +267,30 @@ impl<'clipboard> Set<'clipboard> {
...
+/// Convenience function to get an Objective-C object from a
+/// specific class.
+fn object_class(class: &'static Class) -> Id<NSObject> {
+	// SAFETY: `Class` is valid to transmute into a pointer and the pointer
+	// is a valid object, so its ID can be obtained.
+	let class_id: Id<Class> = unsafe { Id::from_ptr(std::mem::transmute(class)) };
+
+	// SAFETY: All of the variants constructed are valid NSObjects.
+	unsafe { std::mem::transmute(class_id) }
 }
```
kjvalencik, line 296: Why the `transmute` instead of pointer casts?

```suggestion
fn object_class(class: &'static Class) -> Id<NSObject> {
	// SAFETY: `Class` is a valid object and `Id` will not mutate it
	unsafe { Id::from_ptr(class as *const _ as *mut _) }
}
```

src/platform/osx.rs
```diff
@@ -89,17 +91,24 @@ pub(crate) struct Clipboard {
...
-				description: "NSPasteboard#generalPasteboard returned null".into(),
-			});
+
+		if !pasteboard.is_null() {
+			// SAFETY: `generalPasteboard` is not null and a valid object pointer.
+			let pasteboard: Id<Object> = unsafe { Id::from_ptr(pasteboard) };
+			Ok(Clipboard { pasteboard })
+		} else {
+			// Rust only supports 10.7+, while `generalPasteboard` first appeared in 10.0, so this
+			// is unreachable.
```
kjvalencik, line 103: Why not an `unreachable!` if this is the case? What about iPad OS?

src/platform/osx.rs
```diff
@@ -161,102 +170,91 @@ impl<'clipboard> Get<'clipboard> {
...
 	}
 
 	pub(crate) fn text(self, data: String) -> Result<(), Error> {
+		self.clipboard.clear();
+
 		let string_array = NSArray::from_vec(vec![NSString::from_str(&data)]);
-		let _: usize = unsafe { msg_send![self.pasteboard, clearContents] };
-		let success: bool = unsafe { msg_send![self.pasteboard, writeObjects: string_array] };
+		let success: bool =
+			unsafe { msg_send![self.clipboard.pasteboard, writeObjects: string_array] };
```
kjvalencik, line 257: I know the square brackets are the example the `objc` crate gives, but it confusing. What do you think of switching to parenthesis?

src/platform/osx.rs
complexspaces: I need to go back and update this comment, but after pushing this up I found that there are some ways you can run a macOS app (mostly with launch services) that result in the clipboard not being available to the application. With that updated comment in place, I believe the error would then make more sense.

> What about iPad OS?

Currently, nothing in this file will work on iPad OS because it utilizes AppKit. A very similar implementation that uses `UIPasteBoard` would be required for `arboard` to work on iOS and iPad OS.

src/platform/osx.rs
complexspaces: I'm not strongly opinionated one way or another, but I'm not sure how parenthesizes would look with Objective-C call parameters `(writeObjects: string_array`). The brackets also more closely replicate how these calls are written in Objective-C, so it makes referencing examples, etc more straightforward.

src/platform/osx.rs
complexspaces: I believe you're 💯 correct here. The transmutes made more sense in a previous revision of this code which utilized less-typed code but with the current implementation, a pointer cast is much clearer.

src/platform/osx.rs
complexspaces: I have updated this comment to reflect the above.

src/platform/osx.rs
complexspaces: Changed, thanks again for the suggestion.

kjvalencik (approved):

PR #58, opened 2022-06-22
Add support for clearing

complexspaces:
This PR adds support for explicitly clearing the clipboard in the API. For Linux specifically, there is a new extension trait that allows a caller to select which of the Linux clipboards to clear out.

I chose to use another builder for `Clear` to make any future additions easier, along with staying consistent with the existing pattern for Linux extension traits. With that said, it's expected most people will just call `Clipboard::clear_default`, removing some of the concern about type bloat.

This contains changes from #57


src/lib.rs
```diff
@@ -77,6 +77,17 @@ impl Clipboard {
 		self.set().image(image)
 	}
 
+	/// Clears the default clipboard for the platform of any contents that
+	/// may be present.
+	pub fn clear_default(&mut self) -> Result<(), Error> {
+		self.clear().clear()
```
complexspaces, line 83: Note from review: This last clear should be changed to something like `clear_all`.

mathieulj (approved):
Thanks for making this happen!

PR #59, opened 2022-06-23
The scope of the unsafe block can be appropriately reduced

peamaeq:
In this function you use the unsafe keyword for some safe expressions. However, I found that only 6 functions are real unsafe operations (see the list below). 

We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety  within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust. 
**Real unsafe operation list:**
1. the win_to_rgba()\offset()\set_len()\GetDIBits()\CreateDIBitmap()\GetDC() function(these are unsafe functions)

Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html 
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -164,17 +164,17 @@ fn read_cf_dibv5(dibv5: &[u8]) -> Result<ImageData<'static>, Error> {
 		header_size as isize
 	};
 
-	unsafe {
-		let image_bytes = dibv5.as_ptr().offset(pixel_data_start) as *const _;
-		let hdc = GetDC(std::ptr::null_mut());
-		let hbitmap = CreateDIBitmap(
+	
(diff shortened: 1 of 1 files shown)


complexspaces:
I agree with K.J here. There are many subtleties with all of the buffer copying in the image handling so I don't think this is ideal.

Additionally, this user is [apparently a bot](https://github.com/rust-lang/cc-rs/pull/690#issuecomment-1291496453), so I don't think this matters much anyway.

PR #60, opened 2022-07-01
Add WindowsSetExt and improve Windows internals

complexspaces:
This PR does three things:
- Adds a `SetExtWindows` trait
- Reduces the amount of times the Windows clipboard is opened per set of operations
- Makes the Windows clipboard implementation `!Send + !Sync`

Firstly, the addition of the new extension trait is to add support for preventing data set on the Windows clipboard from being synchronized to Microsoft's cloud services or local clipboard history manager. This functionality is desirable in applications that deal with sensitive data, such as as password managers. It would be pretty bad if passwords considered protected ended up in plaintext in OneDrive. Secondly, I refactored the opening of the Windows clipboard to be more consistent with the other platforms by actually opening it when `new()` is called. By keeping the clipboard open between operations, it allows applications to perform multiple operations back-to-back without the worry of being interrupted. 

Finally, I made the `platform::Clipboard` implementation for Windows `!Send + !Sync`. The clipboard on Windows is opened _per thread_, and any operations done on a different thread will always return `ERROR_CLIPBOARD_NOT_OPEN`. To avoid runtime errors, it is now a compile-time error to try and move it between threads. This is a breaking change ~~,but it wouldn't have worked for users on Windows before anyway so  I suspect users may already be dealing with this by using tightly-scoped access to `Clipboard.`~~

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -30,7 +30,7 @@ use winapi::{
 	},
 };
 
-use crate::common::Error;
+use crate::common::{private, Error};
 
 #[cfg(feature = "image-data")]
 use crate::common::{ImageData, ScopeGuard};
@@ -440,11 +440,18 @@ impl<'clipboard> Get<'clipboard> {
 
 pub(crate) struct Set<'clipboard> {
 	clipboard: PhantomData<&'clipboard mut Clipboard>,
+	exclude_from_cloud: bool,
+	exclude_from_history: bool,
 }
 
 impl<'clipboard> Set<'clipboard> {
+	/// `set` should be called with the registered format and a DWORD value of 0.
+	///
+	/// See https://docs.microsoft.com/en-us/windows/win32/dataxchg/clipboard-formats#cloud-clipboard-and-clipboard-history-formats
+	const CLIPBOARD_EXCLUSION_DATA: &'static [u8] = &0u32.to_ne_bytes();
+
 	pub(crate) fn new(_clipboard: &'clipboard mut Clipboard) -> Self {
-		Self { clipboard: PhantomData }
+		Self { clipboard: PhantomData, exclude_from_cloud: false, exclude_from_history: false }
 	}
 
 	pub(crate) fn text(self, data: String) -> Result<(), Error> {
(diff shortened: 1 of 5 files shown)


kjvalencik:
> This is a breaking change, but it wouldn't have worked for users on Windows before anyway

Wouldn't it have worked previously because `.text()` was re-opening the clipboard on each call?

complexspaces:
> Wouldn't it have worked previously because methods like .text() were re-opening the clipboard on each call?

🤦 Thank you for pointing that out. You're 100% correct, sorry about that. In that case, this is just a plain breaking change to better support the new builder patterns available from https://github.com/1Password/arboard/pull/43.

src/platform/windows.rs
```diff
@@ -440,27 +447,52 @@ impl<'clipboard> Get<'clipboard> {
...
+
+		// Clipboard exclusions are applied retroactively to the item that is currently in the clipboard.
+		// See the MS docs on `CLIPBOARD_EXCLUSION_DATA` for specifics. Once the item is added to the clipboard,
+		// tell Windows to remove it from cloud syncing and history.
+
+		if self.exclude_from_cloud {
+			if let Some(format) = clipboard_win::register_format("CanUploadToCloudClipboard") {
+				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
+					.map_err(|_| Error::Unknown {
+						description: "Failed to exclude data from cloud clipboard".into(),
```
mathieulj, line 477: Given that our error enum is now `#[non_exhaustive]`, I wonder if we might be better off adding a specific error variant rather than using `Unknown` here. Yes it's windows specific but users can't exhaustively match anyways so it might not be any worse than `Unknown` containing strings.

complexspaces: What do you think about `FailedToApplyDataProperties` for the variant name? That way it can be more generic and potentially more relevant for macOS additions down the road.

complexspaces: I have added some of this explanation to the callsite for future clarity.

PR #63, opened 2022-07-11
Accept &str instead of String for setting text

complexspaces:
This PR migrates the public and private APIs of the crate to accept borrowed strings instead of owned ones when setting text. I did this for two reasons:
- It better (but not exactly) matches the image setting API, which allows you to use borrowed data.
- Avoiding the overhead of cloning values when not needed on most platforms.

The only platform not helped by this change is Linux with X11. It now requires an internal clone of the data passed in since in. Wayland, and the other OSes, already created "platform specific" owned versions prior.

### DIFF
--- a/src/platform/linux/x11.rs
+++ b/src/platform/linux/x11.rs
@@ -13,6 +13,7 @@ and conditions of the chosen license apply to this file.
 // https://freedesktop.org/wiki/ClipboardManager/
 
 use std::{
+	borrow::Cow,
 	cell::RefCell,
 	collections::{hash_map::Entry, HashMap},
 	sync::{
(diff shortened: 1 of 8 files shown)


src/platform/linux/x11.rs
```diff
@@ -852,12 +852,14 @@ impl Clipboard {
...
 		&self,
-		message: String,
+		message: &str,
 		selection: LinuxClipboardKind,
 		wait: bool,
 	) -> Result<()> {
-		let data =
-			ClipboardData { bytes: message.into_bytes(), format: self.inner.atoms.UTF8_STRING };
+		let data = ClipboardData {
+			bytes: message.as_bytes().to_owned(),
```
kjvalencik, line 860: This line is a bit of a red flag to me. I don't like that `message` is implicitly cloned.

Maybe `set_text` could be generic over `T: Into<Cow<'a, str>> + Clone + 'a` to keep a consistent API across platforms while avoiding `clone` on linux if possible?

kjvalencik (approved):

kjvalencik:
This looks good to me. The downsides:
* More complex API documented (although shouldn't impact usage much)
* Users might not notice that they can remove unnecessary clones (is there a clippy lint for this?)

Overall, I think this is a pretty good balance of avoiding clones when possible and keeping a consistent cross platform API.

kjvalencik:
This looks great @complexspaces!

The only thing that might be valuable is an addition to the doc comment that borrowed data will be cloned on Linux.

PR #66, opened 2022-07-19
Update wl-clipboard-rs crate to 0.6

AldaronLau:


complexspaces:
Thanks for this! It looks `wl-clipboard-rs` made a bunch of updates to clean dependencies up, which is great to see.

complexspaces (approved):

Issue #61, opened 2022-07-04
`content="text/html"` not possible


deggers:
Hey,
I try to use this crate to copy gifs into my clipboard looking like `"<img src=\"data:image/gif;base64,$base64\" >"`.
This works fine when done in Java and put into the `DataFormat.HTML` slot of the `ClipboardContent` but I cannot get this working in rust :( Is it maybe not possible with this crate, because it uses somewhere internally something like `"text/plain"` ?

I use it like the following, not providing a gif because unsure about permssions.
```rust
use arboard::Clipboard;
extern crate image_base64;

fn main() {
	let mut clipboard = Clipboard::new().unwrap();
    
    let image_base64 = image_base64::to_base64("src/pleaseWork.gif");
// Used meta-tag in hope of content negotiation at some higher level 
    let once_more_html = format!(r#"<meta http-equiv="content-type" content="text/html; charset=utf-8"><img alt="" src="{image_base64}">"#);
    clipboard.set_text(once_more_html.to_string()).unwrap()
}
```

complexspaces:
Hey there @deggers. You guessed right that this crate make some assumptions regarding the content type of clipboard contents advertised to different apps on the system. When you use `.set_text`, the contents are advertised as a UTF-8 string (in the platform's most matching system-level content type).

If you're interested, I think that we would accept a PR adding support for the `text/html` format on the [Set builder](https://github.com/1Password/arboard/blob/master/src/lib.rs#L126-L132). It seems like its supported on all the relevant platforms.

complexspaces:
This is now available in the [3.1.1 release](https://crates.io/crates/arboard/3.1.1), which was published to crates.io a few minutes ago.

PR #69, opened 2022-08-02
Implement the `.set_html()` method

breard-r:
This pull request is intended to solve issue #61 by adding new `.set_html()` method that copy HTML content to the clipboard as well as an optional plaintext alternative. The plaintext alternative is useful for programs that do not handle HTML content. Such an interface is already in use on other project like [klembord](https://pypi.org/project/klembord/) (see the `.set_with_rich_text()` method).

## Platforms status

 - [x] Linux (X11)
 - [x] Linux (Wayland)
 - [x] OS X
 - [x] Windows

### DIFF
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -31,8 +31,8 @@ use std::borrow::Cow;
 // Required to bring NSPasteboard into the path of the class-resolver
 #[link(name = "AppKit", kind = "framework")]
 extern "C" {
-	static NSPasteboardTypeHTML: &'static Object;
-	static NSPasteboardTypeString: &'static Object;
+	static NSPasteboardTypeHTML: *const Object;
+	static NSPasteboardTypeString: *const Object;
 }
 
 static NSSTRING_CLASS: Lazy<&Class> = Lazy::new(|| Class::get("NSString").unwrap());
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -273,6 +273,21 @@ impl<'clipboard> Set<'clipboard> {
 
 	pub(crate) fn html(self, html: Cow<'_, str>, alt: Option<Cow<'_, str>>) -> Result<(), Error> {
 		self.clipboard.clear();
+		// Text goes to the clipboard as UTF-8 but may be interpreted as Windows Latin 1.
+		// This wrapping forces it to be interpreted as UTF-8.
+		//
+		// See:
+		// https://bugzilla.mozilla.org/show_bug.cgi?id=466599
+		// https://bugs.chromium.org/p/chromium/issues/detail?id=11957
+		let html = format!(
+			r#"<html>
+				<head>
+					<meta http-equiv="content-type" content="text/html; charset=utf-8">
+				</head>
+				<body>{}</body>
+			</html>"#,
+			html
+		);
 		let html_nss = NSString::from_str(&html);
 		let mut success: bool = unsafe {
 			msg_send![self.clipboard.pasteboard, setString: html_nss forType:NSPasteboardTypeHTML]
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -30,7 +30,10 @@ use std::borrow::Cow;
 
 // Required to bring NSPasteboard into the path of the class-resolver
 #[link(name = "AppKit", kind = "framework")]
-extern "C" {}
+extern "C" {
+	static NSPasteboardTypeHTML: &'static Object;
+	static NSPasteboardTypeString: &'static Object;
+}
 
 static NSSTRING_CLASS: Lazy<&Class> = Lazy::new(|| Class::get("NSString").unwrap());
 #[cfg(feature = "image-data")]
@@ -261,6 +264,27 @@ impl<'clipboard> Set<'clipboard> {
 		let string_array = NSArray::from_vec(vec![NSString::from_str(&data)]);
 		let success: bool =
 			unsafe { msg_send![self.clipboard.pasteboard, writeObjects: string_array] };
+		if success {
+			Ok(())
+		} else {
+			Err(Error::Unknown { description: "NSPasteboard#writeObjects: returned false".into() })
+		}
+	}
+
+	pub(crate) fn html(self, html: Cow<'_, str>, alt: Option<Cow<'_, str>>) -> Result<(), Error> {
+		self.clipboard.clear();
+		let html_nss = NSString::from_str(&html);
+		let mut success: bool = unsafe {
+			msg_send![self.clipboard.pasteboard, setString: html_nss forType:NSPasteboardTypeHTML]
+		};
+		if success {
+			if let Some(alt_text) = alt {
+				let alt_nss = NSString::from_str(&alt_text);
+				success = unsafe {
+					msg_send![self.clipboard.pasteboard, setString: alt_nss forType:NSPasteboardTypeString]
+				};
+			}
+		}
 		if success {
 			Ok(())
 		} else {
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -37,36 +37,6 @@ use crate::common::{ImageData, ScopeGuard};
 
 const MAX_OPEN_ATTEMPTS: usize = 5;
 
-macro_rules! add_clipboard_exclusions {
-    ($obj: ident) => {
-		// Clipboard exclusions are applied retroactively to the item that is currently in the clipboard.
-		// See the MS docs on `CLIPBOARD_EXCLUSION_DATA` for specifics. Once the item is added to the clipboard,
-		// tell Windows to remove it from cloud syncing and history.
-
-		if $obj.exclude_from_cloud {
-			if let Some(format) = clipboard_win::register_format("CanUploadToCloudClipboard") {
-				// We believe that it would be a logic error if this call failed, since we've validated the format is supported,
-				// we still have full ownership of the clipboard and aren't moving it to another thread, and this is a well-documented operation.
-				// Due to these reasons, `Error::Unknown` is used because we never expect the error path to be taken.
-				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
-					.map_err(|_| Error::Unknown {
-						description: "Failed to exclude data from cloud clipboard".into(),
-					})?;
-			}
-		}
-
-		if $obj.exclude_from_history {
-			if let Some(format) = clipboard_win::register_format("CanIncludeInClipboardHistory") {
-				// See above for reasoning about using `Error::Unknown`.
-				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
-					.map_err(|_| Error::Unknown {
-						description: "Failed to exclude data from clipboard history".into(),
-					})?;
-			}
-		}
-    }
-}
-
 #[cfg(feature = "image-data")]
 fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 	use std::intrinsics::copy_nonoverlapping;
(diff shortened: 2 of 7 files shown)


complexspaces:
Hey there @breard-r, thank you for starting on this.

> Unfortunately, I do not have any OS X environment, so I might need help on that platform. That say, I might find someone who is willing to lend me a computer running OS X.

I'm not sure how much time I'll have in the coming weeks, but if you don't get a Mac until then I should be able to help out there, or just act as a tester if you'd prefer to write it yourself.

breard-r:
After nearly a month with very little free time, I might have found a way to get my hand on a MacOS. If everything's ok, I should have access to it within a few weeks, so don't worry if I don't commit anything before the end of september. I haven't forgotten this PR :wink:

ColinKrist:
@breard-r - limited rust experience here - just scoping out a personal project using Tauri to copy discord animated discord emotes into the clipboard to paste into something like MS Teams.

I stumbled on https://madebyevan.com/clipboard-test/ for ripping apart a chrome "Copy Image" event it may be of use for an example for set_html() in the examples:
image

In Windows, I've found that when an app didn't support the GIF copy, it fell back to pasting as an image. I believe the screenshot of the event above shows that, but I'm not exactly sure how this works to have 2 types in a single clipboard event. Is that possible?

breard-r:
@ColinKrist Sorry, I don't think I understand the problem here.

I already implemented `.set_html()` for Windows, it works quite well and it's the app's job to decide what to put in the clipboard. In your example, it's a single `img` tag, nothing more (the other tags as well as the fragment comments are parts of the Windows-specific wrapper, `.set_html()` already takes care of it for you).

If you want to put an image in the clipboard in the same way shown in your screenshot, you can:
```rust
let my_image = "<img href=\"https://example.com/your-image.png\" alt=\"cool image\" />";
let mut ctx = Clipboard::new().unwrap();
ctx.set_html(my_image, None).unwrap();
```

Or maybe you were asking if [.set_image()](https://docs.rs/arboard/2.1.1/arboard/struct.Clipboard.html#method.set_image) could use `.set_html()` as a fallback should it fail?

breard-r:
Status update: today I finally got access to a MacBook Air. I tested something I already coded without being able to test it and, while it stills does not work properly, it's still a good progress and shows that I might be on the right path.

breard-r:
And… it's done! :relieved: 

Time spent finding a MacBook: 2months
Time spent solving the actual problem: a single evening

src/platform/windows.rs
```diff
@@ -37,6 +37,36 @@ use crate::common::{ImageData, ScopeGuard};
 
 const MAX_OPEN_ATTEMPTS: usize = 5;
 
+macro_rules! add_clipboard_exclusions {
```
complexspaces, line 40: I'm not sure about this being a macro over an inherent method on the `Set` builder. It looks like this should work as a method too and be a bit cleaner:
```diff
diff --git a/src/platform/windows.rs b/src/platform/windows.rs
index 6e090f1..76e1575 100644
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -37,36 +37,6 @@ use crate::common::{ImageData, ScopeGuard};
 
 const MAX_OPEN_ATTEMPTS: usize = 5;
 
-macro_rules! add_clipboard_exclusions {
-    ($obj: ident) => {
-		// Clipboard exclusions are applied retroactively to the item that is currently in the clipboard.
-		// See the MS docs on `CLIPBOARD_EXCLUSION_DATA` for specifics. Once the item is added to the clipboard,
-		// tell Windows to remove it from cloud syncing and history.
-
-		if $obj.exclude_from_cloud {
-			if let Some(format) = clipboard_win::register_format("CanUploadToCloudClipboard") {
-				// We believe that it would be a logic error if this call failed, since we've validated the format is supported,
-				// we still have full ownership of the clipboard and aren't moving it to another thread, and this is a well-documented operation.
-				// Due to these reasons, `Error::Unknown` is used because we never expect the error path to be taken.
-				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
-					.map_err(|_| Error::Unknown {
-						description: "Failed to exclude data from cloud clipboard".into(),
-					})?;
-			}
-		}
-
-		if $obj.exclude_from_history {
-			if let Some(format) = clipboard_win::register_format("CanIncludeInClipboardHistory") {
-				// See above for reasoning about using `Error::Unknown`.
-				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
-					.map_err(|_| Error::Unknown {
-						description: "Failed to exclude data from clipboard history".into(),
-					})?;
-			}
-		}
-    }
-}
-
 #[cfg(feature = "image-data")]
 fn add_cf_dibv5(image: ImageData) -> Result<(), Error> {
 	use std::intrinsics::copy_nonoverlapping;
@@ -495,7 +465,7 @@ impl<'clipboard> Set<'clipboard> {
 		clipboard_win::raw::set_string(&data).map_err(|_| Error::Unknown {
 			description: "Could not place the specified text to the clipboard".into(),
 		})?;
-		add_clipboard_exclusions!(self);
+		self.add_clipboard_exclusions()?;
 		Ok(())
 	}
 
@@ -512,7 +482,7 @@ impl<'clipboard> Set<'clipboard> {
 			clipboard_win::raw::set_without_clear(format.get(), &html.into_bytes())
 				.map_err(|e| Error::Unknown { description: e.to_string() })?;
 		}
-		add_clipboard_exclusions!(self);
+		self.add_clipboard_exclusions()?;
 		Ok(())
 	}
 
@@ -526,6 +496,36 @@ impl<'clipboard> Set<'clipboard> {
 
 		add_cf_dibv5(image)
 	}
+
+	fn add_clipboard_exclusions(&self) -> Result<(), Error> {
+		// Clipboard exclusions are applied retroactively to the item that is currently in the clipboard.
+		// See the MS docs on `CLIPBOARD_EXCLUSION_DATA` for specifics. Once the item is added to the clipboard,
+		// tell Windows to remove it from cloud syncing and history.
+
+		if self.exclude_from_cloud {
+			if let Some(format) = clipboard_win::register_format("CanUploadToCloudClipboard") {
+				// We believe that it would be a logic error if this call failed, since we've validated the format is supported,
+				// we still have full ownership of the clipboard and aren't moving it to another thread, and this is a well-documented operation.
+				// Due to these reasons, `Error::Unknown` is used because we never expect the error path to be taken.
+				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
+					.map_err(|_| Error::Unknown {
+						description: "Failed to exclude data from cloud clipboard".into(),
+					})?;
+			}
+		}
+
+		if self.exclude_from_history {
+			if let Some(format) = clipboard_win::register_format("CanIncludeInClipboardHistory") {
+				// See above for reasoning about using `Error::Unknown`.
+				clipboard_win::raw::set_without_clear(format.get(), Self::CLIPBOARD_EXCLUSION_DATA)
+					.map_err(|_| Error::Unknown {
+						description: "Failed to exclude data from clipboard history".into(),
+					})?;
+			}
+		}
+
+		Ok(())
+	}
 }
 
 /// Windows-specific extensions to the [`Set`](super::Set) builder.
```

src/platform/windows.rs
```diff
@@ -465,33 +495,24 @@ impl<'clipboard> Set<'clipboard> {
...
+		let alt = match alt {
+			Some(s) => s.into(),
+			None => String::new(),
+		};
+		clipboard_win::raw::set_string(&alt).map_err(|_| Error::Unknown {
+			description: "Could not place the specified text to the clipboard".into(),
+		})?;
+		if let Some(format) = clipboard_win::register_format("HTML Format") {
+			let html = wrap_html(&html);
+			clipboard_win::raw::set_without_clear(format.get(), &html.into_bytes())
```
complexspaces, line 512: Nit:
```suggestion
			clipboard_win::raw::set_without_clear(format.get(), html.as_bytes())
```

src/platform/osx.rs
```diff
@@ -268,6 +271,27 @@ impl<'clipboard> Set<'clipboard> {
 		}
 	}
 
+	pub(crate) fn html(self, html: Cow<'_, str>, alt: Option<Cow<'_, str>>) -> Result<(), Error> {
+		self.clipboard.clear();
+		let html_nss = NSString::from_str(&html);
+		let mut success: bool = unsafe {
+			msg_send![self.clipboard.pasteboard, setString: html_nss forType:NSPasteboardTypeHTML]
+		};
```
complexspaces, line 279: It looks like both [Chromium](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_mac.mm;l=410-421?q=NSPasteboard) and [Firefox](https://searchfox.org/mozilla-central/source/widget/cocoa/nsClipboard.mm#703-715) wrap the provided HTML into some kind of body, but [both](https://bugs.chromium.org/p/chromium/issues/detail?id=11957) of these [bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=466599) are more than a decade old. Given their age, its very unlikely the issue it was working around still exists but it may be worthwhile to leave a comment for the future saying that these bugs we're at least acknowledged.

complexspaces (changes requested):
This looks like a good start, thank you for taking the time to implement this. I left some comments, but in addition please add a new test cycle that exercises this function and just reads the regular string back.

I'll try and test this locally soon as well.

src/platform/osx.rs
breard-r: Thank you for the tip, I had no idea about this issue. Even if it might be fixed, I think it's safer to wrap the html. It's easy and shouldn't harm since Firefox still uses it (`widget/cocoa/nsClipboard.mm`, `nsClipboard::WrapHtmlForSystemPasteboard`).

complexspaces (approved):
The code here looks pretty good. Thank you for bearing with me and I apologize for not getting to this sooner.

I ran this through some functionality tests on the three platforms and I've found one minor discrepancy. On Ubuntu Linux+X11, the clipboard's HTML properties seem to go away after the clipboard exists, but the regular text sticks around. This seems to be solely be [a limitation of GNOME](https://github.com/oae/gnome-shell-pano/issues/33#issuecomment-1243024554)/clipboard managers, but I'm curious if you can reproduce the behavior:

image

If you'd like, you can visualize it with [this site](https://madebyevan.com/clipboard-test/), or the [Zettlr editor](https://www.zettlr.com/) and see how it reverts to plaintext once `arboard`'s process exits.

breard-r:
> The code here looks pretty good. Thank you for bearing with me and I apologize for not getting to this sooner.

Thank you for the review, it helped me improve! And obviously there is no rush :relaxed: 

> I'm curious if you can reproduce the behavior

Yes: I installed an Ubuntu VM and got the exact same behavior. I can also reproduce it with other software: copying rich text from LibreOffice and pasting it to clipboard-test shows 3 types (`text/plain`, `text/rtf` and `text/html`) if LibreOffice is still opened and only `text/plain` if LibreOffice has been closed before pasting. Therefore, it really seems to a an issue with the clipboard manager itself.

By the way, I guess you already know it, but since other people might read this thread I would like to quote the relevant [X11 clipboard specification](https://specifications.freedesktop.org/clipboards-spec/clipboards-latest.txt) section which also suggest an issue with the clipboard manager:

> A remaining somewhat odd thing about X selections is that exiting the
> app you did a cut/copy from removes the cut/copied data from the
> clipboard, since the selection protocol is asynchronous and requires
> the source app to provide the data at paste time. The solution here
> would be a standardized protocol for a "clipboard daemon" so that apps
> could hand off their data to a daemon when they exit. Or
> alternatively, you can run an application such as xclipboard which
> constantly "harvests" clipboard selections.

On this topic, I learned that some clipboard managers like GPaste actually have [an option to save the rich text or just the plain text](https://github.com/Keruspe/GPaste/blob/v43.0/data/gsettings/org.gnome.GPaste.gschema.xml.in#L126-L132).

complexspaces:
I'll look at releasing a new version this week that will have your changes included. Thanks again for this :)

complexspaces:
This is now released in [3.1.1](https://crates.io/crates/arboard/3.1.1).

PR #70, opened 2022-08-02
Prep to release 3.0.

complexspaces:


kjvalencik:
I am somewhat concerned with the potential for error with the naming of `Clipboard::clear`. The name seems to indicate that it will _clear_ when it only creates a builder.

_**Note:** This does not impact `get` or `set` which more clearly need additional information._

One idea is to rename the method  to `clear_with` to help understanding that it's not performing a clear. The idea is that this would read like English. We can also add a `Clipboard::clear` to reduce the noise of `clipboard.clear_with().clear()`.

```rust
        clipboard
            .clear_with()
            .clipboard(LinuxClipboardKind::Primary)
            .clear()?;
```

> _Clear with_ the primary clipboard.

complexspaces:
I've opened https://github.com/1Password/arboard/pull/71 to look at reworking the public API for clipboard clearing. Please take a look.

PR #85, opened 2022-09-29
Mention Windows changes in the 3.0 release changelog

complexspaces:
I originally missed the breaking changes from #60 when writing the 3.0 release changelog. This PR amends it to mention the clipboard opening changes.


ar37-rs:
@complexspaces hi, thank you for reply. got it. anyway, i'm making my own text editor app for learning pupose and what if there's a feature or similar can be enabled in arboard? so arboard 3.x can behave as arboard 2.x on windows platform, that would be nice and appreciated. thank you.

complexspaces:
@ar37-rs We don't currently have any plans to add a cargo feature to change the behavior. I think that if we changed this around more we would bring back the old behavior and add a builder method that tells it to "lock" the clipboard by keeping it open between operations.

ar37-rs:
@complexspaces, ok never mind. i can still use arboard 2.x works great for me so far, thank you.

Issue #88, opened 2022-10-19
Platform dependent use patterns of `arboard`


emilk:
Since 3.0, we should now create a new `Clipboard`, use it, then drop it right away on Windows.

However, I'm getting reports that that same code doesn't work on Linux X11, and that we there instead need to create a `Clipboard` once and then keep it around for the life of the program: https://github.com/emilk/egui/issues/2138

If true, this makes using `arboard` suddenly very annoying as I need different paths depending on what platform I am compiling for (which is exactly what I want `arboard` to help me abstract).

---

From a users point of view, I just want access to functions to copy and paste text and images. I don't want to think about the lifetime of the `Clipboard` resource.

complexspaces:
Thanks for opening an issue. I agree that this is problematic, and I think that I've now seen it cause enough problems that I'll be reverting the new default behavior introduced in #60 to what it was pre 3.0. Sorry for causing headaches downstream 😦. I did not think enough about how cross-platform users would be affected by the new usage pattern on Windows, which is entirely my bad. I'll be looking at this after work today.

emilk:
Thanks @complexspaces ❤️

derfarctor:
I have to say it is rather unclear from the repository the correct usage pattern for the Clipboard struct. 

Can we expect to be able to (for all platforms and display server protocols) create multiple instances of Clipboard in numerous places in the program and drop them whenever we like without it affecting the underlying data that has either been written to or read from the Clipboard? I.e. create a new Clipboard, set some text, block ends and Clipboard is lost, and expect the text to remain paste-able. 

Or must we create a single instance of Clipboard and use only that when `set_text()`ing or `get_text()`ing for the entirety of the program's lifespan?

complexspaces:
@derfarctor The high-level answer to your question is "no." Linux uses too different of an architecture compared to the other platforms. Do you mind looking through the documentation changes in #89 and letting me know if it would answer your question and if there's anything else you would like to see called out?

derfarctor:
@complexspaces The documentation is more clear there. 

I've changed my program to only use a single instance of Clipboard but now, as I understand is expected, it does not work properly on Windows. I eagerly await the release of the fix such that this usage pattern can be fully cross platform friendly.

complexspaces:
These fixes have been published as `3.2.0` on crates.io. Thanks again for opening this issue and contributing everyone.

PR #89, opened 2022-10-27
Fix Windows clipboard behavior and improve documentation

complexspaces:
This PR fixes the inconsistent behavior introduced by #60 on Windows that made `arboard`'s semantics entirely different on Windows vs the other platforms, preventing reasonable cross-platform usage.

In addition, it improves the documentation of `Clipboard` to note that your app's usage of arboard may differ depending on what combination of platforms are being targeted. For example, if you're only targeting macOS and Windows, nothing beyond opening and using the clipboard in a one-off fashion is needed. However if Linux is used[^1], then one copy of the clipboard should be kept around at all times.

[^1]: The `SetLinuxExt` builder can help with some of these cases, but is not a universal solution for all apps.

Closes https://github.com/1Password/arboard/issues/88

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 3.2.0
+
+### Changed
+- The Windows clipboard now behaves consistently with the other
+platform implementations again.
+
 ## 3.1.1 on 2022-17-10
 
 ### Added
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,9 +8,6 @@ the Apache 2.0 or the MIT license at the licensee's choice. The terms
 and conditions of the chosen license apply to this file.
 */
 
-#![crate_name = "arboard"]
-#![crate_type = "lib"]
-
 mod common;
 use std::borrow::Cow;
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 3.2.0
+
+### Changed
+- The Windows clipboard now behaves consistency with the other
```
emilk, line 6:
```suggestion
- The Windows clipboard now behaves consistently with the other
```

emilk (commented):
Thanks for this 🙏

CHANGELOG.md
complexspaces: Thank you for catching that. It looks like I made the typo in a few places 😅.

ArturKovacs (approved):
Thanks for this PR. I ran into this myself, I'm looking forward to seeing this merged.

complexspaces:
@ArturKovacs This is now released as `3.2.0` 👍

ArturKovacs:
Awesome! ♥

PR #101, opened 2023-03-18
Update `x11rb` to v0.12

daxpedda:
This updates `x11rb` to v0.11.

This increases the MSRV from v1.46 to v1.56. Which should be fine because the MSRV for `image` is also 1.56.

Waiting for a new release of `wl-clipboard-rs` with https://github.com/YaLTeR/wl-clipboard-rs/commit/71b6df79ff95159342a280f6769ee688f800ddec.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,6 +61,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitflags"
+version = "2.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4682ae6287fcf752ecaabbfcc7b6f9b72aa33933dc23a554d853aea8eea8635"
+
+[[package]]
 name = "block"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -83,6 +89,12 @@ name = "byteorder"
 version = "1.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
+
+[[package]]
+name = "cc"
+version = "1.0.79"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
 
 [[package]]
 name = "cfg-if"
@@ -140,7 +152,7 @@ version = "0.22.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2581bbab3b8ffc6fcbd550bf46c355135d16e9ff2a6ea032ad6b9bf1d7efe4fb"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
  "core-foundation",
  "core-graphics-types",
  "foreign-types",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -414,6 +414,19 @@ dependencies = [
  "memoffset",
 ]
 
+[[package]]
+name = "nix"
+version = "0.25.1"
```
complexspaces, line 419: I don't know if this is a blocker, but it is unfortunate that this change would make any user of `arboard` who wanted Wayland functionality stuck with two versions of `nix` by-default.

AFAICT, there is no set of dependency versions that could remove this right now. `wayland-client` never had a version with `nix` 0.25 and `x11rb` has [yet to make a new release](https://crates.io/crates/x11rb/0.11.1/dependencies) which includes `0.26`.

daxpedda: It's a bit unfortunate indeed, I don't mind leaving this until `x11rb` gets an update.

complexspaces: If you don't mind this sitting longer then it already has, that would be my preference as well.

daxpedda: So now that `x11rb` had an update, we still need to update Wayland. Gonna make some PRs to `wl-clipboard-rs`.

daxpedda: Unfortunately I don't know enough about Wayland to do this myself, so I left this here: https://github.com/YaLTeR/wl-clipboard-rs/issues/36.

daxpedda: I know updated `wl-clipboard-rs` with https://github.com/YaLTeR/wl-clipboard-rs/commit/71b6df79ff95159342a280f6769ee688f800ddec.
Changelog is here: https://github.com/YaLTeR/wl-clipboard-rs/blob/master/CHANGELOG.md#unreleased.

This pulled in `windows-sys` v0.48, so I updated `parking_lot_core` to remove the duplicate `windows-sys` dependencies.

There's now only one `nix` version pulled in. So this should be good to go now, just waiting for a new `wl-clipboard-rs` release.

@complexspaces if you could try this PR on Wayland (I don't have a Wayland desktop right now) and report back to https://github.com/YaLTeR/wl-clipboard-rs/issues/36#issuecomment-1604982787 it would be appreciated.

daxpedda: This has been release now.

daxpedda:
This is ready now again.

Unfortunately we have a new duplicate dependency now: `bitflags`.
- https://github.com/image-rs/image-png/pull/400
- https://github.com/servo/core-foundation-rs/issues/633 -> https://github.com/servo/core-foundation-rs/pull/631
- https://gitlab.redox-os.org/redox-os/syscall/-/issues/34
- `wayland-*` needs to upgrade `nix` to v0.27 or move to `rustix`: https://github.com/Smithay/wayland-rs/issues/653 -> https://github.com/Smithay/wayland-rs/pull/658
- `x11rb` needs to release a new version with their move to `rustix`: https://github.com/psychon/x11rb/issues/770 (https://github.com/psychon/x11rb/pull/815)

complexspaces:
Thanks for updating this PR again! Regarding the bitflags update, I don't see it as problematic for merging. Its the first semver incompatible change they've made in a long time and the rest of the crate ecosystem is working through it too.

I will spin up a Wayland VM by the end of the weekend to smoke check this, but otherwise I think this is good to merge once its been rebased for the CI fix you opened earlier.

daxpedda:
> [..], but otherwise I think this is good to merge once its been rebased for the CI fix you opened earlier.

Done.

complexspaces (approved):
I tested this branch in both Wayland via KDE Plasma and then X11 and Wayland on Ubuntu and all tests passed, confirming the updates didn't have any noticeable behavior changes.

This is good to go now, thanks again!

complexspaces:
This is now released in [v3.3.0](https://crates.io/crates/arboard/3.3.0). Thanks for your patience.

PR #102, opened 2023-04-05
fixing doc example so it passes tests on ubuntu

jonZlotnik:
There isn't a CI test for ubuntu, but they should pass locally.
It failed on compile, so the `no_run` wasn't doing anything helpful.

Just added a main as suggested by the rust docs:
https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#using--in-doc-tests

And added "use" for missing dependencies.

### DIFF
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -268,7 +268,7 @@ pub trait ClearExtLinux: private::Sealed {
 	///
 	/// ### Example
 	///
-	/// ```
+	/// ```no_run
 	/// # use arboard::{ Clipboard, LinuxClipboardKind, ClearExtLinux, Error };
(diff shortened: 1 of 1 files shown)


jonZlotnik:
@complexspaces

complexspaces:
Hey there, thanks for the PR and my apologies for taking so long to respond.

While I think that its great to make the doctest compile successfully, we need to keep the `no_run` annotation. This is because the clipboard is a global resource on every OS: Any instance of `arboard::Clipboard` can mess with another. If it was removed, the other tests in the crate would be able to encounter a race condition with this test (depending on when they are run). The other unit tests in `arboard` carefully serialize everything to avoid the problem.

complexspaces (approved):

PR #105, opened 2023-07-24
Remove macOS memory leaks when writing to the clipboard

### LINKED ISSUE #24: Is there a memory leak

complexspaces:
This PR removes the long-outstanding memory leaks in the macOS clipboard backend. After some experimentation, I found that these methods had leaky implementations:
- `set_string`
- `set_image`
- `set_html`

None of the reading functions had leaks AFAICT.

The root cause of (most) the issues was that several calls to `msg_send!` were not passing their arguments correctly. Instead of passing a pointer to the relevant Objective-C object to a class function, they passed the value itself via move. This worked before since the value could still be read by the system, but it became problematic with memory usage. As the system functions only expected a reference to the data, none of them attempted to free it since this would lead to a use-after-free or double-free in user code. So to Rust's eyes, the value was moved across the FFI boundary and never had their destructors called, essentially the same if `mem::forget` had been used.

To resolve these problems, I corrected the callsites to pass a reference/pointer across like the APIs expected.

The other issue was that when creating an `NSImage` class allocation, the previous code was accidentally increasing the reference count after allocating instead of taking ownership. This resulted in an unbalanced retain count that meant the resulting class could never be freed. Allocating a class always has at least one reference, otherwise it would be instantly reclaimed by the runtime. Our user code becomes the object's new owner once the runtime has finished preparing it.

Relates to #24

### DIFF
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -84,10 +84,12 @@ fn image_from_pixels(
 		kCGRenderingIntentDefault,
 	);
 	let size = NSSize { width: width as CGFloat, height: height as CGFloat };
-	let image: Id<NSObject> = unsafe { Id::from_ptr(msg_send![*NSIMAGE_CLASS, alloc]) };
+	// Take ownership of the newly allocated object, which has an existing retain count.
+	let image: Id<NSObject> = unsafe { Id::from_retained_ptr(msg_send![*NSIMAGE_CLASS, alloc]) };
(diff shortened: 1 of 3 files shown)


complexspaces:
This PR does not fix all of the reported issues like I originally hoped, but these are still valid leaks that should be fixed, so I'm going to merge this in parallel with investigating the getter leaks.

HuakunShen:
Thank you for your work! @complexspaces

PR #110, opened 2023-08-24
For consistency use png on macOS

VZout:
Instead of storing as png on linux and tiff on macos, it feels nicer to just store it as png everywhere.

(haven't tested it yet, but macos docs say this is possible)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ edition = "2018"
 
 [features]
 default = ["image-data"]
-image-data = ["core-graphics", "image", "winapi/minwindef", "winapi/wingdi", "winapi/winnt"]
-wayland-data-control = ["wl-clipboard-rs"]
+image-data = ["dep:core-graphics", "dep:image", "winapi/minwindef", "winapi/wingdi", "winapi/winnt"]
+wayland-data-control = ["dep:wl-clipboard-rs"]
 
 [dependencies]
 thiserror = "1.0"
@@ -36,7 +36,7 @@ objc_id = "0.1"
 objc-foundation = "0.1"
 once_cell = "1"
 core-graphics = { version = "0.22", optional = true }
-image = { version = "0.24", optional = true, default-features = false, features = ["tiff"] }
+image = { version = "0.24", optional = true, default-features = false, features = ["png"] }
 
 [target.'cfg(all(unix, not(any(target_os="macos", target_os="android", target_os="emscripten"))))'.dependencies]
 log = "0.4"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -11,8 +11,8 @@ edition = "2018"
 
 [features]
 default = ["image-data"]
-image-data = ["core-graphics", "image", "winapi/minwindef", "winapi/wingdi", "winapi/winnt"]
-wayland-data-control = ["wl-clipboard-rs"]
+image-data = ["dep:core-graphics", "dep:image", "winapi/minwindef", "winapi/wingdi", "winapi/winnt"]
```
VZout, line 14: This makes the feature flags in the docs a bit more readable as right now it shows `image`, `core-graphics` and `wl-clipboard-rs` as "features" but they are just optional dependencies

complexspaces:
Thanks for the PR. I think this change makes sense as long as it doesn't break copying compatibility in any noticeable way. Could I ask you to try some testing by copying images from various apps/websites to see if there's a difference in what `NSPasteboard` can convert into a `.png` vs a `.tiff`?

Cargo.toml
complexspaces: FWIW, this requires a bump to Rust 1.60 to support:
>  The dep: syntax is only available starting with Rust 1.60. Previous versions can only use the implicit feature name.

`arboard` doesn't currently have an MSRV, and 1.60 is over a year old at this point, so I don't think its a big deal. Its also easy enough to rollback if it causes someone problems. I'm mentioning it just in case though.

src/platform/osx.rs
```diff
@@ -227,14 +227,14 @@ impl<'clipboard> Get<'clipboard> {
 			Some(_) | None => return Err(Error::ContentNotAvailable),
 		};
 
-		let tiff: &NSArray<NSObject> = unsafe { msg_send![obj, TIFFRepresentation] };
+		let png: &NSArray<NSObject> = unsafe { msg_send![obj, PNGRepresentation] };
```
complexspaces, line 230: I don't believe NSImage supports this representation (and is causing [a test failure](https://github.com/1Password/arboard/actions/runs/5962987149/job/16202321208?pr=110)), [unlike UIImage which does](https://developer.apple.com/documentation/uikit/1624096-uiimagepngrepresentation?language=objc).

One alternative may be to wait for https://github.com/1Password/arboard/pull/106 to merge which switches to a different method of getting images from the clipboard. That way [does support PNG](https://developer.apple.com/documentation/appkit/nspasteboardtypepng?language=objc) like you'd want.

complexspaces (changes requested):
Regarding my comment about compatibility testing, I got a little bit ahead of myself since I didn't look at the code first. Sorry. Feel free to play around on my other PR though where you could fit PNG reading in though.

complexspaces:
Closing in favor of https://github.com/1Password/arboard/pull/164 or future PRs doing similar.

PR #119, opened 2023-11-21
Make winapi crate optional

rhysd:
This PR makes winapi crate dependency optional.

Context:

- My application doesn't handle image data. So I use this crate without `image-data` features.
- winapi crate is very inactive (no commit for 3 years). Avoiding it as much as possible would be a good idea.

Without `image-data`, `winapi` is used only for `syncapi::Sleep` in this crate and it can be easily replaced with Rust standard `std::thread::sleep`. `std::thread::sleep` [uses the same `Sleep` C function](https://github.com/rust-lang/rust/blob/3a85a5cfe7884f94e3cb29a606913d7989ad9b48/library/std/src/sys/windows/thread.rs#L100) under the hood. As a side effect, `std::thread::sleep` leverages high-resolution timer. It could make the retry interval more accurate.

This doesn't remove winapi dependency immediately since [clipboard-win](https://crates.io/crates/clipboard-win) v4 still depends on it. However clipboard-win will [replace winapi dependency with windows-rs](https://github.com/DoumanAsh/clipboard-win/commit/ff1a6dd54392d41aa239b6eafc5ed28f9703606b) at next major version v5. When this crate bumps clipboard-win to v5, winapi dependency can be removed along with this PR (without `image-data` feature).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ thiserror = "1.0"
 env_logger = "0.9.0"
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.9", features = [
+winapi = { version = "0.3.9", optional = true, features = [
     "basetsd",
     "winuser",
     "winbase",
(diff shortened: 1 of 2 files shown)


complexspaces (approved):
Thank you for the PR, this seems reasonable to me.

Regarding `winapi` being used for image functionality, once a new version of `clipboard-win` is out I can replace `winapi` with `windows-sys` here, to remove it from the dependency tree fully.

PR #126, opened 2023-12-29
Spell check docs

robjtede:
quick cSpell run


complexspaces (approved):
Thanks for the cleanup.

PR #127, opened 2023-12-29
Document errors on Clipboard

robjtede:


src/lib.rs
```diff
@@ -55,32 +55,49 @@ pub use platform::SetExtWindows;
 ///
 /// The clipboard on Windows is a global object, which may only be opened on one thread at once.
 /// This means that `arboard` only truly opens the clipboard during each operation to ensure that
-/// multiple `Clipboard`'s may exist at once. This also means that attempting operations in parallel
-/// has a high likelyhood to return an error instead.
+/// multiple `Clipboard`s may not exist at once (on the same thread). This also means that
```
complexspaces, line 58: On Windows this is a system-wide limitation (so for `arboard`'s documentation, per process), not per-thread. The lack of `Send+Sync` there is a lint to try and stop you from getting into this state when assuming other apps are behaving right.

src/common.rs
```diff
@@ -46,11 +46,8 @@ pub enum Error {
 	#[error("The native clipboard is not accessible due to being held by an other party.")]
 	ClipboardOccupied,
 
-	/// This can happen in either of the following cases.
-	///
-	/// - When returned from `set_image`: the image going to the clipboard cannot be converted to the appropriate format.
-	/// - When returned from `get_image`: the image coming from the clipboard could not be converted into the `ImageData` struct.
-	/// - When returned from `get_text`: the text coming from the clipboard is not valid UTF-8 or cannot be converted to UTF-8.
+	/// The image or the text that was about the be transferred to/from the clipboard could not be
+	/// converted to the appropriate format.
```
robjtede, line 50: these cases are better documented on the methods themselves; remove from here to prevent de-sync

complexspaces (changes requested):
Do you mind squashing/rebasing on `main` as well? I can also handle it if you would rather not but I try not to use merge commits.

src/lib.rs
robjtede: I think we agree on the problem here but maybe this tweak is not very good. How do you want this to read? Maybe an explicit call out to the user that they should avoid creating/operating on clipboard objects on >1 threads? That's not something that the !Send+!Sync can prevent.

robjtede:
> Do you mind squashing/rebasing on main as well

sure, can do

> I try not to use merge commits

does this repo not allow squash merging?

src/lib.rs
complexspaces: > Maybe an explicit call out to the user that they should avoid creating/operating on clipboard objects on >1 threads?

That sounds good to me. Its likely easier for the user to reason about not creating more then 1 across >1 thread then creating N clipboards and manually synchronizing them.

> That's not something that the !Send+!Sync can prevent.

I know it isn't, but it does help with the Windows requirement of per-thread clipboarding. If you send the clipboard object to another thread, it won't behave correctly.

robjtede: Thoughts on the new wording? @complexspaces

src/lib.rs
```diff
@@ -54,33 +54,52 @@ pub use platform::SetExtWindows;
 /// ## Windows
 ///
 /// The clipboard on Windows is a global object, which may only be opened on one thread at once.
-/// This means that `arboard` only truly opens the clipboard during each operation to ensure that
-/// multiple `Clipboard`'s may exist at once. This also means that attempting operations in parallel
-/// has a high likelihood to return an error instead.
+/// This means that `arboard` only truly opens the clipboard during each operation to prevent
+/// multiple `Clipboard`s from existing at once.
+///
+/// This means that attempting operations in parallel has a high likelihood to return an error. As
```
complexspaces, line 60:
```suggestion
/// This means that attempting operations in parallel has a high likelihood to return an error or deadlock. As
```

src/lib.rs
```diff
@@ -54,33 +54,52 @@ pub use platform::SetExtWindows;
...
 }
 
 impl Clipboard {
-	/// Creates an instance of the clipboard
+	/// Creates an instance of the clipboard.
+	///
+	/// # Errors
+	///
+	/// On some platforms or desktop environments, an error can be returned if clipboards are not
+	/// supported.
```
complexspaces, line 73:
```suggestion
	/// supported. This may be retried. 
```

src/lib.rs
```diff
@@ -54,33 +54,52 @@ pub use platform::SetExtWindows;
...
 	pub fn new() -> Result<Self, Error> {
 		Ok(Clipboard { platform: platform::Clipboard::new()? })
 	}
 
 	/// Fetches UTF-8 text from the clipboard and returns it.
+	///
+	/// # Errors
+	///
+	/// Returns error if clipboard is empty or contents is not UTF-8 text.
```
complexspaces, line 82:
```suggestion
	/// Returns error if clipboard is empty or contents are not UTF-8 text.
```

src/lib.rs
```diff
@@ -94,6 +113,11 @@ impl Clipboard {
 	/// Any image data placed on the clipboard with `set_image` will be possible read back, using
 	/// this function. However it's of not guaranteed that an image placed on the clipboard by any
 	/// other application will be of a supported format.
+	///
+	/// # Errors
+	///
+	/// Returns error if clipboard is empty, contents is not an image, or contents cannot be
```
complexspaces, line 119:
```suggestion
	/// Returns error if clipboard is empty, contents are not an image, or the contents cannot be
```

complexspaces (changes requested):

src/lib.rs
complexspaces: Sorry, I missed the notification for the most recent push to the branch. I left one more nit on this phrasing but otherwise I think it looks good to go 👍

complexspaces (approved):
Thanks!

PR #130, opened 2024-01-19
Implement getting HTML on X11

lucieleblanc:


src/common.rs
```diff
@@ -146,6 +146,22 @@ impl<'a> ImageData<'a> {
...
+}
+
+impl HTMLData {
+	pub fn from_html(html: String) -> Self {
+		Self { html, ..Default::default() }
+	}
+
+	pub fn from_alt_text(alt_text: String) -> Self {
+		Self { alt_text, ..Default::default() }
+	}
```
complexspaces, line 162: Is this a valid function to expose? If the user doesn't have any HTML in the first place it seems like they'd be better off using `set_text()` instead.

src/common.rs
```diff
@@ -146,6 +146,22 @@ impl<'a> ImageData<'a> {
 	}
 }
 
+#[derive(Debug, Default)]
```
complexspaces, line 149: Nit: Let's remove the `Default` implementation here. It makes it harder to maintain version compatibility in the future.

src/lib.rs
```diff
@@ -78,6 +79,11 @@ impl Clipboard {
 		self.set().text(text)
 	}
 
+	/// Fetches HTML from the clipboard and returns it.
+	pub fn get_html(&mut self) -> Result<HTMLData, Error> {
+		self.get().html()
+	}
```
complexspaces, line 85: Question: Do you think HTML is a common enough case to add this to the main `Clipboard` structure? I would prefer to expose this only through the `Get` builder to keep it tidy.

src/platform/linux/x11.rs
```diff
@@ -884,6 +887,39 @@ impl Clipboard {
...
+			self.inner.atoms.STRING,
+			self.inner.atoms.TEXT,
+			self.inner.atoms.TEXT_MIME_UNKNOWN,
+		];
+
+		let mut data = HTMLData::default();
+
+		let html_result = self.inner.read(&html_format, selection);
+		if let Ok(html_data) = html_result {
+			let html: String = html_data.bytes.into_iter().map(|c| c as char).collect();
```
complexspaces, line 905: Is this correct? My understanding that HTML can contain UTF-8 so we should just use `String::from_utf8` here.

src/platform/linux/x11.rs
```diff
@@ -884,6 +887,39 @@ impl Clipboard {
...
+	pub(crate) fn get_html(&self, selection: LinuxClipboardKind) -> Result<HTMLData> {
+		let html_format = [self.inner.atoms.HTML];
+		let alt_text_formats = [
+			self.inner.atoms.UTF8_STRING,
+			self.inner.atoms.UTF8_MIME_0,
+			self.inner.atoms.UTF8_MIME_1,
+			self.inner.atoms.STRING,
+			self.inner.atoms.TEXT,
+			self.inner.atoms.TEXT_MIME_UNKNOWN,
+		];
```
complexspaces, line 899: Can we find a way to extract this into a reusable construct that `get_text` can use as well?

src/platform/linux/x11.rs
```diff
@@ -884,6 +887,39 @@ impl Clipboard {
...
+		let alt_text_result = self.inner.read(&alt_text_formats, selection);
+		if let Ok(alt_text_data) = alt_text_result {
+			data.alt_text = if alt_text_data.format == self.inner.atoms.STRING {
+				// ISO Latin-1
+				// See: https://stackoverflow.com/questions/28169745/what-are-the-options-to-convert-iso-8859-1-latin-1-to-a-string-utf-8
+				alt_text_data.bytes.into_iter().map(|c| c as char).collect::<String>()
+			} else {
+				String::from_utf8(alt_text_data.bytes).map_err(|_| Error::ConversionFailure)?
+			};
+		}
```
complexspaces, line 918: Same here, because the logic is also duplicated from `get_text`.

complexspaces (changes requested):
Thanks for the PR! I didn't look at this originally because it was still a draft but I had a spare moment and thought it may be helpful to guide this one in.

complexspaces:
With @Gae24's help another version of this was merged via https://github.com/1Password/arboard/pull/163.

I'm going to close this as superseded, but thank you for your initial change skeleton. If you run into any limitations, please open a followup issue.

PR #134, opened 2024-02-22
Add `SetExtLinux#wait_until`

auguwu:
Hello :wave:!

I added a `wait_until` method on [`SetExtLinux`](https://github.com/1Password/arboard/blob/77e0e078eb460ac2fa0eda96124163c11ef6b2d1/src/platform/linux/mod.rs#L181) for X11 to wait until a deadline was exceeded so that:

* You wouldn't need a daemon process to call `wait` on.
* Easier for CLI programs who don't want to infinitely loop / wait for new contents to be replaced & want to finish quickly.

This does report the following log when used:
```sh
2024-02-22T18:28:30.515424Z  WARN arboard::platform::linux::x11: Could not hand the clipboard contents over to the clipboard manager. The request timed out.
```

since `result.timed_out` is used and I'm not too sure what to do in this case:
https://github.com/1Password/arboard/blob/77e0e078eb460ac2fa0eda96124163c11ef6b2d1/src/platform/linux/x11.rs#L711-L715

### DIFF
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -154,9 +154,15 @@ impl<'clipboard> Set<'clipboard> {
 
 	pub(crate) fn text(self, text: Cow<'_, str>) -> Result<(), Error> {
 		match self.clipboard {
-			Clipboard::X11(clipboard) => {
-				clipboard.set_text(text, self.selection, self.wait, self.wait_until)
-			}
+			Clipboard::X11(clipboard) => clipboard.set_text(
+				text,
+				self.selection,
+				match (self.wait, self.wait_until) {
+					(_, Some(deadline)) => x11::WaitConfig::Until(deadline),
+					(true, None) => x11::WaitConfig::Forever,
+					(false, None) => x11::WaitConfig::None,
+				},
+			),
 
 			#[cfg(feature = "wayland-data-control")]
 			Clipboard::WlDataControl(clipboard) => clipboard.set_text(text, self.selection, self.wait),
@@ -165,9 +171,16 @@ impl<'clipboard> Set<'clipboard> {
 
 	pub(crate) fn html(self, html: Cow<'_, str>, alt: Option<Cow<'_, str>>) -> Result<(), Error> {
 		match self.clipboard {
-			Clipboard::X11(clipboard) => {
-				clipboard.set_html(html, alt, self.selection, self.wait, self.wait_until)
-			}
+			Clipboard::X11(clipboard) => clipboard.set_html(
+				html,
+				alt,
+				self.selection,
+				match (self.wait, self.wait_until) {
+					(_, Some(deadline)) => x11::WaitConfig::Until(deadline),
+					(true, None) => x11::WaitConfig::Forever,
+					(false, None) => x11::WaitConfig::None,
+				},
+			),
 
 			#[cfg(feature = "wayland-data-control")]
 			Clipboard::WlDataControl(clipboard) => clipboard.set_html(html, alt, self.selection, self.wait),
@@ -177,9 +190,15 @@ impl<'clipboard> Set<'clipboard> {
 	#[cfg(feature = "image-data")]
 	pub(crate) fn image(self, image: ImageData<'_>) -> Result<(), Error> {
 		match self.clipboard {
-			Clipboard::X11(clipboard) => {
-				clipboard.set_image(image, self.selection, self.wait, self.wait_until)
-			}
+			Clipboard::X11(clipboard) => clipboard.set_image(
+				image,
+				self.selection,
+				match (self.wait, self.wait_until) {
+					(_, Some(deadline)) => x11::WaitConfig::Until(deadline),
+					(true, None) => x11::WaitConfig::Forever,
+					(false, None) => x11::WaitConfig::None,
+				},
+			),
 
 			#[cfg(feature = "wayland-data-control")]
 			Clipboard::WlDataControl(clipboard) => clipboard.set_image(image, self.selection, self.wait),
@@ -241,13 +260,11 @@ pub trait SetExtLinux: private::Sealed {
 	/// Whether or not to wait for the clipboard's content to be replaced after setting it. This waits until the
 	/// `deadline` has exceeded.
 	///
-	/// This is useful for short-lived programs so that it doesn't block until new contents on the clipboard
-	/// were added and will exit as so.
+	/// This is useful for short-lived programs so it won't block until new contents on the clipboard
+	/// were added.
 	///
-	/// For X11, this will wait until it either had new contents available in the clipboard or if the
-	/// `deadline` was exceeded. This isn't available for Wayland and will not do anything.
-	///
-	/// Note: this will call [`wait()`][SetExtLinux::wait] if it wasn't previously set.
+	/// Note: this is a superset of [`wait()`][SetExtLinux::wait] and will overwrite any state
+	/// that was previously set using it.
 	fn wait_until(self, deadline: Instant) -> Self;
 }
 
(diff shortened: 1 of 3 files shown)


src/platform/linux/mod.rs
```diff
@@ -143,25 +143,32 @@ impl GetExtLinux for crate::Get<'_> {
 pub(crate) struct Set<'clipboard> {
 	clipboard: &'clipboard mut Clipboard,
 	wait: bool,
+	wait_until: Option<Instant>,
```
complexspaces, line 146: Suggestion: I believe this would be better if we refactored these away into a single enum representing the possible cases:
```rust
enum WaitConfig {
    None,
    Forever,
    Until(Instant)
}
```

It would also reduce the amount of parameters in the various `set` functions and simplify the state handling via `match`.

src/platform/linux/mod.rs
```diff
@@ -227,6 +237,18 @@ pub trait SetExtLinux: private::Sealed {
...
+	/// `deadline` has exceeded.
+	///
+	/// This is useful for short-lived programs so that it doesn't block until new contents on the clipboard
+	/// were added and will exit as so.
+	///
+	/// For X11, this will wait until it either had new contents available in the clipboard or if the
+	/// `deadline` was exceeded. This isn't available for Wayland and will not do anything.
+	///
+	/// Note: this will call [`wait()`][SetExtLinux::wait] if it wasn't previously set.
+	fn wait_until(self, deadline: Instant) -> Self;
```
complexspaces, line 251: Please move this underneath `wait` so the related methods are grouped together.

src/platform/linux/mod.rs
```diff
@@ -227,6 +237,18 @@ pub trait SetExtLinux: private::Sealed {
...
+	/// Whether or not to wait for the clipboard's content to be replaced after setting it. This waits until the
+	/// `deadline` has exceeded.
+	///
+	/// This is useful for short-lived programs so that it doesn't block until new contents on the clipboard
+	/// were added and will exit as so.
+	///
+	/// For X11, this will wait until it either had new contents available in the clipboard or if the
+	/// `deadline` was exceeded. This isn't available for Wayland and will not do anything.
+	///
+	/// Note: this will call [`wait()`][SetExtLinux::wait] if it wasn't previously set.
```
complexspaces, line 250:
```suggestion
	/// Note: this is a superset of [`wait()`][SetExtLinux::wait] and will overwrite any state that was previously set using it.
```

src/platform/linux/mod.rs
```diff
@@ -227,6 +237,18 @@ pub trait SetExtLinux: private::Sealed {
 	/// # }
 	/// ```
 	fn clipboard(self, selection: LinuxClipboardKind) -> Self;
+
+	/// Whether or not to wait for the clipboard's content to be replaced after setting it. This waits until the
+	/// `deadline` has exceeded.
+	///
+	/// This is useful for short-lived programs so that it doesn't block until new contents on the clipboard
+	/// were added and will exit as so.
```
complexspaces, line 245:
```suggestion
	/// This is useful for short-lived programs so that it doesn't block until new contents on the clipboard
	/// were added and will exit automatically once the specified deadline has elapsed.
```

complexspaces (changes requested):
I apologize for the review delay on this PR. I think this is reasonably useful functionality given how weird Linux can be, and it doesn't conflict with anything else.

auguwu:
I can do that as soon as I can. Also, is the WARN log message ok to be emitted or can it be removed?

complexspaces:
As far as I know, the `warn` log is working as intended and unrelated to your changes here. Do you have a clipboard manager on your system? And are you able to share the code you used to test this?

auguwu:
I was able to test it and it does work correctly when I made this PR. I'll test it again and see if it still works once I can fix the suggested changes.

I'm not sure how to do unit tests for this though so I might have to dig through how the 'wait' function does its tests

src/platform/linux/mod.rs
auguwu: I'm not sure which section that the contents should go into

auguwu:
> And are you able to share the code you used to test this?

I am using it for the purposes of [`auguwu/ume`](https://github.com/auguwu/ume), which uses arboard to copy a link when taking a screenshot (`ume screenshot`). It really boils down to this:

```rust
// the code that does the copying
let mut clipboard = arboard::Clipboard::new()?;
cfg_if::cfg_if! {
  if #[cfg(target_os = "linux")] {
      use arboard::SetExtLinux;

      // wait ~1s to block
      clipboard.set().wait_until(Instant::now() + Duration::from_secs(1)).text(url)?;
    } else {
      clipboard.set_text(url)?;
    }
  }
}
```

and this is how the application reacts to it:
```console
$ UME_UPLOADER_KEY="1234" cargo run -- screenshot http://localhost:3621
2024-04-10T01:30:11.430927Z  INFO ume::cli::cmds::screenshot: found `flameshot` program :3 flameshot=/run/current-system/sw/bin/flameshot
2024-04-10T01:30:11.431155Z  INFO ume::cli::cmds::screenshot: creating file... file=/tmp/screenshots/2024-04-09T18:30:11.430982407-07:00.png
2024-04-10T01:30:11.432066Z  INFO ume::cli::cmds::screenshot: $ /run/current-system/sw/bin/flameshot gui -r > /tmp/screenshots/2024-04-09T18:30:11.430982407-07:00.png
2024-04-10T01:30:17.069959Z  INFO ume::cli::cmds::screenshot: uploading file to Ume server... file=/tmp/screenshots/2024-04-09T18:30:11.430982407-07:00.png
2024-04-10T01:30:17.070032Z  INFO ume::cli::cmds::screenshot: Now uploading file... file=/tmp/screenshots/2024-04-09T18:30:11.430982407-07:00.png
2024-04-10T01:30:17.149519Z  INFO ume::cli::cmds::screenshot: copying url to clipboard!
2024-04-10T01:30:18.150255Z  INFO ume::cli::cmds::screenshot: copied to clipboard, deleting image
2024-04-10T01:30:18.250519Z  WARN arboard::platform::linux::x11: Could not hand the clipboard contents over to the clipboard manager. The request timed out.
```

and it'll copy `http://localhost:3621/images/{random characters}.png` to the clipboard.

src/platform/linux/mod.rs
complexspaces: Sorry, I meant to indicate that this this function (and its documentation) should be moved above `fn clipboard` but below `fn wait` in the trait definition.

src/platform/linux/mod.rs
complexspaces: @auguwu Your last changeset brought this most of the way, but the fields inside `struct Set` should also get replaced with the new enum. You'll need to bring `WaitConfig` up into `mod.rs` as well and update the `wayland` module to handle the `Forever` case.

The `fn wait` and `fn wait_until` methods should then look something like this:
```rust
fn wait(mut self) -> Self {
    self.platform.wait = WaitConfig::Forever;
    self
}

fn wait_until(mut self, deadline: Instant) -> Self {
    self.platform.wait = WaitConfig::Until(deadline);
    self
}
```

src/platform/linux/x11.rs
```diff
@@ -205,6 +205,20 @@ enum ReadSelNotifyResult {
 	EventNotRecognized,
 }
 
+/// Configuration on how long to wait for a new X11 copy event is emitted.
+#[derive(Default)]
+pub enum WaitConfig {
```
complexspaces, line 210:
```suggestion
pub(crate) enum WaitConfig {
```

auguwu:
I should fix everything that was suggested!

complexspaces (approved):

complexspaces:
Thanks again for the PR, and additionally thank you for the `TRACE` logs using `arboard` on your system. It seems like something is amiss in the clipboard ordering handoff but that can happen in another pull request/issue for sure.

LOHAYBOY999 (approved):

Issue #136, opened 2024-02-29
v3.3.1 broken on windows


ardyfeb:
Cannot compile on windows with latest version `v3.3.1`

```
error[E0446]: private type `platform::windows::OpenClipboard<'_>` in public interface
   --> C:\Users\gitlab_runner\.cargo\registry\src\index.crates.io-6f17d22bba15001f\arboard-3.3.1\src\platform\windows.rs:40:2
    |
40  |     pub fn add_cf_dibv5(_open_clipboard: OpenClipboard, image: ImageData) -> Result<(), Error> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private type
...
422 | struct OpenClipboard<'clipboard> {
    | -------------------------------- `platform::windows::OpenClipboard<'_>` declared as private
For more information about this error, try `rustc --explain E0446`.
```

haakonfp:
Seems to be caused by the refactor in a100f2d77c5b2877257220e9144179d1b78784fe, where the `add_cf_dibv5` function is moved into the new `mod image_data`. The mod has `use super::*` but `OpenClipboard` is private, so it can't access it.

This can be fixed by simply adding `pub` to the `struct OpenClipboard`. However, that might compromise the intended API design.
It looks like the `_open_clipboard` is unused in the function, and only there to enforce the existence of an open clipboard? At least that is my impression by seeing it's usage later in the file (line #600).

I can open a PR for adding `pub` to the `OpenClipboard` struct, just in case that's a sufficient fix.
@rhysd, maybe you have input on a better solution?

complexspaces:
Hi everyone, thanks for the report.

Can I ask what `cargo` command can be used to reproduce this? Somehow it slipped past CI and my manual testing so I'd like to add a regression test to ensure future issues which have visibility mishaps are caught ahead of time. I can't seem to reproduce this yet with `cargo build --features image-data`.

complexspaces:
I am pretty certain https://github.com/1Password/arboard/pull/138 fixes this problem, but could I ask someone who can actually reproduce this issue to use the `fix-windows-image-visiblity` branch in their build and confirm this?

Once this is confirmed to be the fix (and ideally I can reproduce it), I'll make a new patch release fixing the problem.

haakonfp:
[...]

PR #137, opened 2024-03-01
Fix build error on Windows by making `OpenClipboard` public

haakonfp:
Ref issue #136.

Fixes build error on Windows.
Might want another solution for this that doesn't directly expose the `OpenClipboard` type.

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -419,7 +419,7 @@ impl Drop for Clipboard {
 	fn drop(&mut self) {}
 }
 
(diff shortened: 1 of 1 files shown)


complexspaces:
Thank you for the eager contribution to fix the issue! I opened up an alternate PR via #138 that also has some hardening against similar issues cropping up in the future. And, like you said, `OpenClipboard` should not be public because its an internal detail of how `arboard` manages clipboard locking on Windows where that's needed.

I'm going to go ahead and close this as superseded, but thanks again for the attempt.

Issue #135, opened 2024-02-26
Preserving image transparency on Windows


Klavionik:
Hello! 

I'm authoring `copykitten`, a Python library that provides clipboard operations utilizing `arboard` under the hood. I have an issue report from a user, who has a problem pasting transparent images into PowerPoint on Windows: https://github.com/Klavionik/copykitten/issues/5. I was able to confirm this issue and now looking for a solution.

There's one way, which is to make `arboard` put images into the clipboard not only in bitmap and DIB formats but also as a PNG file. This is a trick used by Google Chrome, for example. I believe this would be useful for many Windows applications, not only PowerPoint. 

I have a working draft of code that does that and as far as I can see it works fine. If you are interested in this I could work on a pull request implementing this feature.

complexspaces:
Hey @Klavionik,

At first glance I think this would be fine. Would you be able to detail about what this would entail? After looking around for a bit I see:
- `BITMAPV5HEADER` supports setting `bV5Compression`  to `BI_PNG`. With that set, a number of other fields are basically just passing in PNG metadata like height, width, etc.
- This could switch based on if the user has copied a format with transparency (like PNG) vs raw RBG(A) bytes. This could be trivially detected by looking at magic bytes.

Klavionik:
Hey!

I suppose this could be done by having a method similar to `add_cf_dibv5`, but with a different set of steps. Then these methods will be called in `Set::image` method of `windows` module, like this.

```rust
pub(crate) fn image(self, image: ImageData) -> Result<(), Error> {
    // skipped...
    
    image_data::add_cf_dibv5(open_clipboard, image)
    image_data::add_png(image) // This is the new method.
}
```

This will also require having `image` library on Windows, so that we can take raw RGBA bytes and save it as PNG to the memory buffer before copying it.

This wouldn't require any changes to the public API, as far as I see.

Klavionik:
I suppose I can close this issue now, as this is fixed.

PR #141, opened 2024-03-04
Copy image as PNG file on Windows

Klavionik:
@complexspaces Hello! I decided to just go ahead and open a PR regarding the issue #135. Hope you don't mind.

I'm pretty new to Rust, so pardon me if I did something plain wrong here. :) Just wanted to help you understand the idea that I proposed in more detail. 

Here's how it looks if you paste a transparent image into PowerPoint now:

transparent

And this is how it looked before:

opaque

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ windows-sys = { version = "0.48.0", optional = true, features = [
 ]}
 clipboard-win = "5.3.1"
 log = "0.4"
+image = { version = "0.25", optional = true, default-features = false, features = ["png"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc = "0.2"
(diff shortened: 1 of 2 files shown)


Klavionik:
@complexspaces Just a kind reminder! I would be glad if you could find some time to review this.

complexspaces (approved):
Thank you for taking the time to make a PR! I took a bit too long to finish looking over this one, but I wanted to better understand how more of the Windows ecosystem handles transparent images. After looking through Firefox, Java, Chromium, etc I've convinced myself this is the right approach for better compatibility.

One thing I noticed is that other apps use `DIBV5` plus `BI_RGB` as a secondary option to PNG encoding data. This is something `arboard` can look at later but isn't a blocker for this PR.

I took a look through your changes and only had a few minor codestyle comments. This needed rebased on `main` as well to pick up the recent `image` 0.25 update too. In order to save some time, since I already took a while to get around to reviewing this, I updated the branch myself with my suggestions.

Klavionik:
Glad to hear it's merged! Thank you too, for taking care of the PR.

PR #148, opened 2024-04-23
Use `objc2` and its framework crates

madsmtm:
[`objc2`](https://github.com/madsmtm/objc2) is a replacement for `objc`/`objc_id` that contains a bunch of safety improvements, including `msg_send_id!` which automatically upholds memory management rules (`Id::from_ptr`/`Id::from_retained_ptr` is no longer necessary).

Additionally, we use the framework crates `objc2-foundation` and `objc2-app-kit`, which provide for example the `NSPasteboard` type, which has the methods that arboard needs already defined, and with the correct types, ensuring that passing e.g. `Id<NSArray>` and thus accidentally giving away ownership over the array won't happen again.

These crates are automatically generated, ensuring that if you need some obscure API in the future, it's very likely to be there already.

Feel free to ask if there's anything about the new code that's unclear!

### DIFF
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -11,31 +11,18 @@ and conditions of the chosen license apply to this file.
 use crate::common::Error;
 #[cfg(feature = "image-data")]
 use crate::common::ImageData;
-#[cfg(feature = "image-data")]
-use core_graphics::{
-	base::{kCGBitmapByteOrderDefault, kCGImageAlphaLast, kCGRenderingIntentDefault, CGFloat},
-	color_space::CGColorSpace,
-	data_provider::{CGDataProvider, CustomData},
-	image::CGImage,
+use objc2::{
+	msg_send_id,
+	rc::{autoreleasepool, Id},
+	runtime::ProtocolObject,
+	ClassType,
 };
-use objc::{
-	msg_send,
-	rc::autoreleasepool,
-	runtime::{Class, Object},
-	sel, sel_impl,
+use objc2_app_kit::{NSPasteboard, NSPasteboardTypeHTML, NSPasteboardTypeString};
+use objc2_foundation::{NSArray, NSString};
+use std::{
+	borrow::Cow,
+	panic::{RefUnwindSafe, UnwindSafe},
 };
-use objc_foundation::{INSArray, INSFastEnumeration, INSString, NSArray, NSObject, NSString};
-use objc_id::{Id, Owned};
-use std::{borrow::Cow, ptr::NonNull};
-
-// Required to bring NSPasteboard into the path of the class-resolver
-#[link(name = "AppKit", kind = "framework")]
-extern "C" {
-	static NSPasteboardTypeHTML: *const Object;
-	static NSPasteboardTypeString: *const Object;
-	#[cfg(feature = "image-data")]
-	static NSPasteboardTypeTIFF: *const Object;
-}
 
 /// Returns an NSImage object on success.
 #[cfg(feature = "image-data")]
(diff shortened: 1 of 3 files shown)


src/platform/osx.rs
```diff
@@ -81,42 +71,54 @@ fn image_from_pixels(
...
+	let cg_image: *const c_void = cg_image.cast();
+
 	let size = NSSize { width: width as CGFloat, height: height as CGFloat };
-	let nsimage_class = objc::class!(NSImage);
-	// Take ownership of the newly allocated object, which has an existing retain count.
-	let image: Id<NSObject> = unsafe { Id::from_retained_ptr(msg_send![nsimage_class, alloc]) };
-	#[allow(clippy::let_unit_value)]
-	{
-		// Note: `initWithCGImage` expects a reference (`CGImageRef`), not an actual object.
-		let _: () = unsafe { msg_send![image, initWithCGImage: &*cg_image size:size] };
```
madsmtm, line 91: This line was UB, `initWithCGImage:size:` returns that image object, not `void`.

src/platform/osx.rs
```diff
@@ -81,42 +71,54 @@ fn image_from_pixels(
...
 	Ok(image)
 }
 
 pub(crate) struct Clipboard {
-	pasteboard: Id<Object>,
+	pasteboard: Id<NSPasteboard>,
 }
 
+unsafe impl Send for Clipboard {}
+unsafe impl Sync for Clipboard {}
```
madsmtm, line 95: These are... Probably not entirely true, `NSPasteboard` is likely only safe to use from the main thread, as that's where other code in an application would be using them from - at least, the thread-safety of `NSPasteboard` isn't documented anywhere.

The impls are added to keep backwards compatibility.

src/platform/osx.rs
```diff
@@ -81,42 +71,54 @@ fn image_from_pixels(
...
 	pub(crate) fn new() -> Result<Clipboard, Error> {
-		let cls = Class::get("NSPasteboard").expect("NSPasteboard not registered");
-		let pasteboard: *mut Object = unsafe { msg_send![cls, generalPasteboard] };
+		// Rust only supports 10.7+, while `generalPasteboard` first appeared
+		// in 10.0, so this should always be available.
+		//
+		// However, in some edge cases, like running under launchd (in some
+		// modes) as a daemon, the clipboard object may be unavailable, and
+		// then `generalPasteboard` will return NULL even though it's
+		// documented not to.
```
madsmtm, line 107: I only adapted the comment from before, don't actually know if this is true.

src/platform/osx.rs
```diff
@@ -261,11 +245,9 @@ impl<'clipboard> Set<'clipboard> {
 	pub(crate) fn text(self, data: Cow<'_, str>) -> Result<(), Error> {
 		self.clipboard.clear();
 
-		let string_array = NSArray::from_vec(vec![NSString::from_str(&data)]);
-		// Make sure that we pass a pointer to the system and not the array object itself. Otherwise,
-		// the system won't free it because the API doesn't give it ownership of the data. This results in
-		// a memory leak because Rust can never run its destructor.
-		let success = unsafe { msg_send![self.clipboard.pasteboard, writeObjects: &*string_array] };
+		let string_array =
+			NSArray::from_vec(vec![ProtocolObject::from_id(NSString::from_str(&data))]);
```
complexspaces, line 249: Is `NSArray::from_slice` any better to use here and for the image array below?

complexspaces (commented):
Thank you for this PR. Its great to see that `objc2` is making progress from when I last looked at it. I have some small comments, but otherwise this looks good.

src/platform/osx.rs
madsmtm: It would be, but it isn't possible because the `ProtocolObject<dyn NSPasteboardWriting>` that we end up creating cannot be safely retained in the current `objc2`, so we have to promise (which we can do with the `Vec`) that we're passing ownership to the `NSArray`.

This will be fixed as part of / after https://github.com/madsmtm/objc2/issues/563.

src/platform/osx.rs
complexspaces: Great catch, thank you.

src/platform/osx.rs
complexspaces: FWIW we have been using `NSPasteboard` off the main thread (inside Tokio workers and `std::thread::spawn`) at 1Password for several years now and haven't seen a single crash with a clipboard-related stacktrace. This of course isn't a guarantee but at minimum it hasn't been problematic.

My intuition says this would be somewhat thread-safe because its just an IPC/XPC wrapper to talk to the `pboard` daemon, and XPC uses queues underneath.

src/platform/osx.rs
complexspaces: I don't remember where I read this anymore unfortunately. I should have written it down at the start. This `NUL` handling is harmless but I'm happy to remove the comment if someone eventually proves it inaccurate. `launchd` services are painful enough to deal with I don't want to try them all out today.

src/platform/osx.rs
complexspaces: Got it, thank you for the helpful clarification.

complexspaces (approved):
Thank you again for these changes!

PR #155, opened 2024-06-09
Add Set::exclude_from_history on linux by setting x-kde-passwordManagerHint

### LINKED ISSUE #129: `exclude_from_history` not implemented for KDE klipper clipboard history
I am using the bitwarden desktop app on Linux/KDE, which uses arboard to manage the clipboard and noticed that passwords are stored in the KDE Klipper clipboard manager history.

This can be circumvented by adding `x-kde-passwordManagerHint` MIME data hint to the clipboard entry and setting it to `secret`, see:

https://bugs.kde.org/show_bug.cgi?id=458063

This is implemented in keepassxreboot in this PR:

https://github.com/keepassxreboot/keepassxc/pull/1969/commits

---
BTW the current source code of keepassxreboot also contains an implementation that prevents macosx to put secrets into the clipboard manager: https://github.com/keepassxreboot/keepassxc/blob/7bfe9065cf3a7b9e32094d965eb13a8c7fc0cc82/src/gui/Clipboard.cpp#L56 So maybe this should be part of the generic interface.

MrSmoer:
Fixes #129
Implements the Set::exclude_from_history method for linux.

I tested this on Plasma Wayland (they implement data control protocol) and X11 using this short rust program:
```rust
use arboard::Clipboard;
#[cfg(target_os = "linux")]
use arboard::SetExtLinux;
use std::{thread, time};

fn main() {
    let ten_millis = time::Duration::from_secs(10);
    let mut clipboard = Clipboard::new().unwrap();

    let d = clipboard.set().exclude_from_history().text("secretext");
    println!("hidden");
    thread::sleep(ten_millis);
    let d = clipboard.set().text("notsecrettext");
    println!("Not hidden");
    thread::sleep(ten_millis);
}
```
It did not display the `secretext` when pressing `SUPER+V`(opens klipper) but it did display the `notsecrettext` after 10 seconds, so it works as intended.

Gnome doesn't implement the data control protocol, but testing this change with `wl-paste -l`(probably falling back to the xserver clipboard stuff) seemed to work as well.

In contrast, using a gnome-shell-extension that hooks the the [get_mimetypes-method](https://gjs-docs.gnome.org/st14~14/st.clipboard#method-get_mimetypes) from within the compositor for determinig the clipboard-data-content didn't seem to work properly.
With KeepassXC, which is just using Qt's-functions for handling the clipboard, this function behaved as expected on gnome from what I can tell.

I believe there is some deeper issue with the way gnome handles the clipboard in general, but I was not able to track down the specific difference between how Qt and arboard register their clipboard content to the compositor yet.

Thanks for reviewing this and have a nice Day!

### DIFF
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -8,6 +8,10 @@ use percent_encoding::percent_decode_str;
 use crate::ImageData;
 use crate::{common::private, Error};
 
+// Magic strings used in `Set::exclude_from_history()` on linux
+const KDE_EXCLUSION_MIME: &str = "x-kde-passwordManagerHint";
+const KDE_EXCLUSION_HINT: &[u8] = b"secret";
+
 mod x11;
 
 #[cfg(feature = "wayland-data-control")]
@@ -184,38 +188,56 @@ pub(crate) struct Set<'clipboard> {
 	clipboard: &'clipboard mut Clipboard,
 	wait: WaitConfig,
 	selection: LinuxClipboardKind,
+	exclude_from_history: bool,
 }
 
 impl<'clipboard> Set<'clipboard> {
 	pub(crate) fn new(clipboard: &'clipboard mut Clipboard) -> Self {
-		Self { clipboard, wait: WaitConfig::default(), selection: LinuxClipboardKind::Clipboard }
+		Self {
+			clipboard,
+			wait: WaitConfig::default(),
+			selection: LinuxClipboardKind::Clipboard,
+			exclude_from_history: false,
+		}
 	}
 
 	pub(crate) fn text(self, text: Cow<'_, str>) -> Result<(), Error> {
 		match self.clipboard {
-			Clipboard::X11(clipboard) => clipboard.set_text(text, self.selection, self.wait),
+			Clipboard::X11(clipboard) => {
+				clipboard.set_text(text, self.selection, self.wait, self.exclude_from_history)
+			}
 
 			#[cfg(feature = "wayland-data-control")]
-			Clipboard::WlDataControl(clipboard) => clipboard.set_text(text, self.selection, self.wait),
+			Clipboard::WlDataControl(clipboard) => {
+				clipboard.set_text(text, self.selection, self.wait, self.exclude_from_history)
+			}
 		}
 	}
 
 	pub(crate) fn html(self, html: Cow<'_, str>, alt: Option<Cow<'_, str>>) -> Result<(), Error> {
 		match self.clipboard {
-			Clipboard::X11(clipboard) => clipboard.set_html(html, alt, self.selection, self.wait),
+			Clipboard::X11(clipboard) => {
+				clipboard.set_html(html, alt, self.selection, self.wait, self.exclude_from_history)
+			}
 
 			#[cfg(feature = "wayland-data-control")]
-			Clipboard::WlDataControl(clipboard) => clipboard.set_html(html, alt, self.selection, self.wait),
+			Clipboard::WlDataControl(clipboard) => {
+				clipboard.set_html(html, alt, self.selection, self.wait, self.exclude_from_history)
+			}
 		}
 	}
 
 	#[cfg(feature = "image-data")]
 	pub(crate) fn image(self, image: ImageData<'_>) -> Result<(), Error> {
 		match self.clipboard {
-			Clipboard::X11(clipboard) => clipboard.set_image(image, self.selection, self.wait),
+			Clipboard::X11(clipboard) => {
+				clipboard.set_image(image, self.selection, self.wait, self.exclude_from_history)
+			}
 
 			#[cfg(feature = "wayland-data-control")]
-			Clipboard::WlDataControl(clipboard) => clipboard.set_image(image, self.selection, self.wait),
+			Clipboard::WlDataControl(clipboard) => {
+				clipboard.set_image(image, self.selection, self.wait, self.exclude_from_history)
+			}
 		}
 	}
 }
@@ -280,6 +302,13 @@ pub trait SetExtLinux: private::Sealed {
 	/// # }
 	/// ```
 	fn clipboard(self, selection: LinuxClipboardKind) -> Self;
+
+	/// Excludes the data which will be set on the clipboard from being added to
+	/// the desktop clipboard managers' histories by adding the MIME-Type `x-kde-passwordMangagerHint`
+	/// to the clipboard's selection data.
+	///
+	/// This is the most widely adopted convention on Linux.
+	fn exclude_from_history(self) -> Self;
 }
 
 impl SetExtLinux for crate::Set<'_> {
@@ -295,6 +324,11 @@ impl SetExtLinux for crate::Set<'_> {
 
 	fn wait_until(mut self, deadline: Instant) -> Self {
 		self.platform.wait = WaitConfig::Until(deadline);
+		self
+	}
+
+	fn exclude_from_history(mut self) -> Self {
+		self.platform.exclude_from_history = true;
 		self
 	}
 }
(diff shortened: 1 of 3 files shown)


src/platform/linux/mod.rs
```diff
@@ -254,6 +255,14 @@ pub trait SetExtLinux: private::Sealed {
 	/// # }
 	/// ```
 	fn clipboard(self, selection: LinuxClipboardKind) -> Self;
+
+
+	/// Excludes the data which will be set on the clipboard from being added to
+	/// the cliobard managers' histories by adding the MIME-Type x-kde-passwordMangagerHint
+	/// with the content 'secret' to the clipboard.
```
complexspaces, line 262:
```suggestion
	/// the desktop clipboard managers' histories by adding the MIME-Type `x-kde-passwordMangagerHint`
	/// to the clipboard's selection data.
```

src/platform/linux/mod.rs
```diff
@@ -254,6 +255,14 @@ pub trait SetExtLinux: private::Sealed {
...
+	///
+	/// This is seems to be the convention for this on Linux
```
complexspaces, line 264:
```suggestion
	/// This is the most widely adopted convention on Linux.
```

src/platform/linux/wayland.rs
```diff
@@ -79,15 +79,31 @@ impl Clipboard {
...
 		opts.clipboard(selection.try_into()?);
 		let source = Source::Bytes(text.into_owned().into_bytes().into_boxed_slice());
-		opts.copy(source, MimeType::Text).map_err(|e| match e {
-			CopyError::PrimarySelectionUnsupported => Error::ClipboardNotSupported,
-			other => into_unknown(other),
-		})?;
+		if exclude_from_history {
+
+			let password_manager_hint = Source::Bytes(b"secret".to_vec().into());
+			let password_manager_hint_mime = MimeType::Specific(String::from("x-kde-passwordManagerHint"));
```
complexspaces, line 91: Suggestion: Can we move these magic strings into `const` values and then then do something like this to make it more clear these aren't dynamic:
```rust
MimeSource { source: Source::Bytes(Box::new(KDE_EXCLUSION_HINT)), mime_type: MimeType::Specific(String::from(KDE_EXCLUSION_MIME) }
```

src/platform/linux/wayland.rs
```diff
@@ -79,15 +79,31 @@ impl Clipboard {
...
+
+			opts.copy_multi(vec![
+				MimeSource { source: source, mime_type: MimeType::Text },
+				MimeSource { source: password_manager_hint, mime_type: password_manager_hint_mime },
+			]).map_err(|e| match e {
+				CopyError::PrimarySelectionUnsupported => Error::ClipboardNotSupported,
+				other => into_unknown(other),
+			})?;
```
complexspaces, line 99: Suggestion: You should be able to move this `.map_err` block outside of the `if {} else {}` to avoid duplication since the logic is the same.

src/platform/linux/x11.rs
```diff
@@ -878,11 +879,20 @@ impl Clipboard {
...
-		let data = vec![ClipboardData {
+		let mut data = vec![];
+		data.push(ClipboardData {
 			bytes: message.into_owned().into_bytes(),
 			format: self.inner.atoms.UTF8_STRING,
-		}];
+		});
+		if exclude_from_history {
+			data.push(ClipboardData {
+				bytes: b"secret".to_vec(),
```
complexspaces, line 891: Same here with the `secret` constant. Maybe it should live in `/linux/mod.rs` for both implementations to use?

complexspaces (changes requested):
Thank you for this PR! The structure of it looks good. My suggestions are just minor code quality ones.

I would like to test this out for myself (probably tomorrow) but I once this is cleaned up I think its ready.

MrSmoer:
I just noticed that it didn't work properly for me without `wayland-data-control` on plasma and sway in any version.

... and I am not sure what to do about the CI-failure

complexspaces:
> I just noticed that it didn't work properly for me without wayland-data-control on plasma-wayland and sway in any version. It works properly with the option enabled.

>  on plasma-wayland and sway in any version. It works properly with the option enabled.

IIRC Plasma defaults to Wayland these days. Do you know if `xwayland` is running or installed? That would be needed for anything to happen if you aren't using the `wayland-data-control` feature. Regarding Sway, the same thing applies. I can certainly try to test this on Ubuntu where `xwayland` comes by default.

> ... and I am not sure what to do about the CI-failure

This looks like a new lint from the latest stable Rust release. At the top of the `x11.rs` file, inside the `use std` block, you'll need to delete `usize` from the list.

MrSmoer:
I tried a little, and I wasn't able to properly solve the new CI-Issue

complexspaces:
I spent several hours tonight trying to test this out, with mixed results. This seems to work find in Plasma Wayland like you observed, but GNOME/Ubuntu were a mess. In my testing there, I used [pano](https://github.com/oae/gnome-shell-pano) as a clipboard monitor. 

GNOME apparently doesn't support the wayland extension that `wl-clipboard-rs` [uses](https://github.com/YaLTeR/wl-clipboard-rs/issues/39) to put data on the clipboard, so `arboard` doesn't support wayland on GNOME for the time being, I guess. I believe that a proper-windowed app like KeePassXC uses a different protocol that actually works. I don't know off the top of my head what is is though.

Next testing in a GNOME X11 environment, I still couldn't get it to work with this PR. KeePassXC works fine. From what I can see, it _works while the process using `arboard` is still running_ but once the `arboard`-specific X11 Window has been destroyed (when `Clipboard` is dropped) the clipboard manager saves the contents for some reason. I wasn't able to figure out the reason.

I'm not sure how to proceed here. It doesn't feel right to merge this when it can't be confirmed working-as-intended. I most likely will not have the time to continue digging into the X11 behavior in-depth because there's so many moving parts. Do you have interest in doing so @MrSmoer?

MrSmoer:
I already started looking into the way the clipboard data flows, but could really wrap my head around it. The whole stack on gnome seems very convoluted to me, especially in connection with Xwayland and stuff. I dont have thaaat much time to spare, so this is going to be on my list of side-projects. I am not sure what would have to change in the ecosystem but I believe that it's just the way arboard interfaces with the x11-package or within the x11 rust-package. It would take some time :)

complexspaces:
Hey again! I'm following up here with the intent to merge this branch now that I've resolved the last outstanding issue from my prior testing. I was doing some other clipboard stuff the last few weeks and ended up getting a lot of helpful context to investigate what was happening here. I apologize that this sat for so long but I didn't have any motivation or additional knowledge to revisit this until last week.

> ... I believe that a proper-windowed app like KeePassXC uses a different protocol that actually works. I don't know off the top of my head what is is though.

To quickly document this: GUI applications utilize the `wl_data_device` protocol for their clipboard operations, which is very distinct from what `wl-clipboard-rs` uses. GNOME has no intent to support the `ext-data-manager` extensions though, so my previous statement about GNOME incompatibility is going to be accepted as-is.

> ... From what I can see, it works while the process using arboard is still running but once the arboard-specific X11 Window has been destroyed (when Clipboard is dropped) the clipboard manager saves the contents for some reason. I wasn't able to figure out the reason.

This is what I spent more time digging into. After some testing, I narrowed it down to the way we were destroying the X Window when the last `Clipboard` is dropped. KeePassXC non-deterministically also displayed similar behavior if its process was suddenly terminated without the chance to cleanup. If `arboard`'s window is destroyed while we still own the data, some X servers see that as a crash or something and forcibly retain the current selection's string contents. 

I added a proper ownership relinquishment call and all the issues went away, so the GNOME clipboard manager extensions are no longer retaining excluded values. I decided to continue not doing this for non-excluded values though for two reasons:
1. Generally people really want these saved and get annoyed when things disappear, so this is another possible chance for the clipboard manager to do its job.
2. I don't want to risk a behavior regression for existing users right now 😅.

With that I believe we have proper coverage of X11 and Wayland clipboard manager exclusions now and am happy to merge this. For transparency I swapped around the commit authors on this branch due to the amount of time and debugging I put into the functionality. You're still obviously a co-author though because of the original contribution set.

complexspaces (approved):
Thanks again for starting on this! I also added support for excluding the other `Set` types and tested them on X11 with success.

complexspaces:
This is now released as part of [3.6.0](https://crates.io/crates/arboard/3.6.0). Thanks again!

PR #157, opened 2024-07-11
Add Get::html on macOS (unimplemented on others)

fasterthanlime:
So this is half-baked since it panics on non-macOS platforms, but I can work on finishing it up if that's something y'all are willing to bear the burden of maintaining long-term.

If not, it'll just live in a fork of mine! Let me know!

(also just noticing https://github.com/1Password/arboard/pull/130 now)

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -176,7 +176,7 @@ impl<F: FnOnce()> Drop for ScopeGuard<F> {
 pub(crate) mod private {
 	// This is currently unused on macOS, so silence the warning which appears
 	// since there's no extension traits making use of this trait sealing structure.
-	#[cfg_attr(target_vendor = "apple", allow(unreachable_pub))]
+	#[cfg_attr(target_vendor = "apple", allow(unreachable_pub, dead_code))]
 	pub trait Sealed {}
 
 	impl Sealed for crate::Get<'_> {}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,6 +179,11 @@ impl Get<'_> {
 		self.platform.text()
 	}
 
+	/// Completes the "get" operation by fetching HTML from the clipboard.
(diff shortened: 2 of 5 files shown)


complexspaces:
Hey there, thanks for the PR. HTML is something we've committed to having it on the [platform agnostic main API](https://docs.rs/arboard/latest/arboard/struct.Clipboard.html#method.set_html), so we'd be happy to wrap up that dangling thread by adding a matching `get` method to the crate. All of our supported platforms today (and future ones) can handle it just fine so I don't foresee any painful long-term maintenance 👍.

> (also just noticing https://github.com/1Password/arboard/pull/130 now)

Yeah, its been a few months since that one had any activity. I think that my comment from that PR about where this API should live still feels valid, that its probably best to leave this functionality on the `Get` structure for now and consider adding it to the main clipboard API later if there's demand since its a non-breaking addition.

This crate doesn't have an explicit CLA, but I don't think it would be problematic to cherry-pick that PRs commit here and improve on it in more commits for merge if that's something that interests you vs writing X11 from scratch.

fasterthanlime:
I’m not sure how much time I will have to dedicate to this, but for what it’s worth I’m totally fine with this pull request being cannibalized by any further pull request that tackles all four target platforms at once!

Gae24:
Hi, I was looking to get this done primarily on Linux, but I can also delve into how to implement it for other platforms too.
So isn't it a problem if I include your changes @fasterthanlime ?

fasterthanlime:
> Hi, I was looking to get this done primarily on Linux, but I can also delve into how to implement it for other platforms too. So isn't it a problem if I include your changes @fasterthanlime ?

No problem at all, please do!

complexspaces:
With @Gae24's help this was merged via #163.

I'm going to close this as superseded, but thank you for your initial work!

PR #160, opened 2024-08-17
Add support for wasm32 + web

DouglasDwyer:
This PR implements the ability to cut, copy, and paste text on the web/WASM platforms. This allows libraries like `egui-winit` to use the clipboard on the web (see https://github.com/emilk/egui/issues/4270). This is accomplished in the following way:

- Setting or clearing the clipboard is accomplished by calling `window.navigator.clipboard.write()`
- Reading from the clipboard is accomplished by adding a `paste` event listener to the root HTML document, which stores the `event.clipboardData.getData("text")` value into a global variable. Whenever the clipboard contents are requested from Rust, the global variable is read. _Note that this means the user must actually perform a paste before the clipboard contents become visible to Rust. This is a caveat, but it's the only way that I could find to implement synchronous pasting._
- Image copying/pasting is not implemented. When images are pasted in a browser, they are returned in the `event.clipboardData.files` array, where `async` is required to read the data.


src/platform/wasm.rs
```diff
@@ -0,0 +1,133 @@
...
+        Ok(Self {
+            inner,
+            _paste_callback: paste_callback,
+            window
+        })
+    }
+
+    fn get_last_clipboard(&self) -> String {
+        js_sys::Reflect::get(&self.window, &Self::GLOBAL_CLIPBOARD_OBJECT.into())
+            .ok().and_then(|x| x.as_string()).unwrap_or_default()
```
complexspaces, line 41: AFAICT this return an `Option` so that the `get` functions can return the proper `ContentNotAvailable` error when nothing (or anything in the right format) is present.

src/platform/wasm.rs
```diff
@@ -0,0 +1,133 @@
...
+	clipboard: &'clipboard mut Clipboard,
+}
+
+impl<'clipboard> Clear<'clipboard> {
+	pub(crate) fn new(clipboard: &'clipboard mut Clipboard) -> Self {
+		Self { clipboard }
+	}
+
+	pub(crate) fn clear(self) -> Result<(), Error> {
+        let _ = self.clipboard.inner.write(&js_sys::Array::default());
```
complexspaces, line 60: Should the various clipboard calls failing return `Error::Unknown`?

src/platform/wasm.rs
```diff
@@ -0,0 +1,133 @@
...
+        Ok(())
+	}
+
+	pub(crate) fn html(self, html: Cow<'_, str>, alt: Option<Cow<'_, str>>) -> Result<(), Error> {
+		let alt = match alt {
+			Some(s) => s.into(),
+			None => String::new(),
+		};
+        
+        self.clipboard.set_last_clipboard(&html);
```
complexspaces, line 108: Should this be moved down closer to the `self.clipboard.inner.write` call for a lower chance of things getting out of sync? It also makes it easier to see this function both updates the custom global and DOM clipboard correctly.

complexspaces (changes requested):
Thank you for the PR. I think the approach here makes sense for typical web applications. I'm not a web developer though 😅.

Question: Do you have the ability to test this in Safari? I know their Clipboard APIs are a bit different compared to Firefox and Chromium. I have the ability to test it myself but would need an example WASM app to run.

src/platform/wasm.rs
```diff
@@ -0,0 +1,133 @@
...
+    inner: web_sys::Clipboard,
+    window: web_sys::Window,
+    _paste_callback: web_sys::wasm_bindgen::closure::Closure<dyn FnMut(web_sys::ClipboardEvent)>
+}
+
+impl Clipboard {
+    const GLOBAL_CLIPBOARD_OBJECT: &str = "__arboard_global_clipboard";
+
+    pub(crate) fn new() -> Result<Self, Error> {
+        let window = web_sys::window().ok_or(Error::ClipboardNotSupported)?;
```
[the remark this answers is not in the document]
DouglasDwyer: Unfortunately web workers cannot be supported - the clipboard API is not available in web workers (see the note on [this page](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API)). I will add some documentation to mention this.

src/platform/wasm.rs
DouglasDwyer: The calls with `let _ = ` don't fail immediately; they return a `Promise` that resolves once the browser has updated the clipboard contents. Since this is a synchronous API, I just fire and forget the promise here.

The only instances in which these calls fail (according to [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations)) is if the page doesn't have permission to write to the user's clipboard. Clipboard permissions are granted with transient activation - that is, once the user interacts with the page, writing to the clipboard is allowed.

It would be nice if there was a way to report the error here. The fire and forget approach is a bit jank - but it *does* work for my use case, which is just handling normal copy/paste interactions with `egui` on the web.

src/platform/wasm.rs
```diff
@@ -0,0 +1,133 @@
...
+        let inner = window.navigator().clipboard();
+        
+        let window_clone = window.clone();
+        let paste_callback = web_sys::wasm_bindgen::closure::Closure::wrap(Box::new(move |e: web_sys::ClipboardEvent| {
+            if let Some(data_transfer) = e.clipboard_data() {
+                js_sys::Reflect::set(&window_clone, &Self::GLOBAL_CLIPBOARD_OBJECT.into(), &data_transfer.get_data("text").unwrap_or_default().into())
+                    .expect("Failed to set global clipboard object.");
```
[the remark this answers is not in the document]
DouglasDwyer: This is part of the closure, so it's not possible to bail by returning an error. However, I have changed the code so that it sets the `GLOBAL_CLIPBOARD_OBJECT` to null if `data_transfer.get_data("text")` doesn't work. This will cause subsequent clipboard reads to return `Err(ContentNotAvailable)`.

DouglasDwyer:
I've gone ahead and attempted to address your feedback! The web has some security restrictions on what's possible, but this PR works perfectly for me with `egui` on the web, so I appreciate your review.

> Question: Do you have the ability to test this in Safari? I know their Clipboard APIs are a bit different compared to Firefox and Chromium. I have the ability to test it myself but would need an example WASM app to run.

I sadly don't have access to a Mac right now. However, I've put together a really simple [example web app](https://github.com/DouglasDwyer/arboard-example) for testing. You should be able to just clone the repository and run `trunk serve` to test it.

complexspaces:
Hi, just a quick update: I don't have the bandwidth at the moment to finish up the testing and reviewing for this PR. I apologize for the late communication but I'll need to circle back to it later, likely in October.

complexspaces:
Hello again, sorry for the radio silence. I ended up getting bogged down trying to figure out how to test this cross-platform and ensure consistent functionality.

https://github.com/1Password/arboard/issues/191 describes unblocking this being merged, so I would like to ask for slightly more time to set that up. My intent is that this would sit no longer then summer.

atollk:
@complexspaces Are there still plans to get this merged?

Its-Just-Nans:
Just FYI

using egui I did https://github.com/Its-Just-Nans/bladvak/blob/85ec632b8d78b1635f460b2ddd46d13242b85f0e/src/utils/clipboard.rs#L196-L266

https://docs.rs/bladvak/latest/bladvak/utils/clipboard/struct.BladvakClipBoard.html

with poll_promise, it can handle text and images


Example usage at https://github.com/Its-Just-Nans/wombat/blob/5aacd0afc222f2a130558b7de503c726fcbb37b9/src/windows/importer/mod.rs#L175-L219

PR #164, opened 2024-10-14
macOS: switch to png for image

Gae24:
Used png for handling images on all platform. On macOS dropped the dependency on core-graphics since now an image is directly stored as png

### DIFF
--- a/src/platform/osx.rs
+++ b/src/platform/osx.rs
@@ -207,18 +207,18 @@ impl<'clipboard> Get<'clipboard> {
 
 	#[cfg(feature = "image-data")]
 	pub(crate) fn image(self) -> Result<ImageData<'static>, Error> {
-		use objc2_app_kit::NSPasteboardTypeTIFF;
+		use objc2_app_kit::NSPasteboardTypePNG;
 		use std::io::Cursor;
 
 		// XXX: There does not appear to be an alternative for obtaining images without the need for
 		// autorelease behavior.
 		let image = autoreleasepool(|_| {
-			let image_data = unsafe { self.clipboard.pasteboard.dataForType(NSPasteboardTypeTIFF) }
+			let image_data = unsafe { self.clipboard.pasteboard.dataForType(NSPasteboardTypePNG) }
 				.ok_or(Error::ContentNotAvailable)?;
 
 			let data = Cursor::new(image_data.bytes());
 
-			let reader = image::io::Reader::with_format(data, image::ImageFormat::Tiff);
+			let reader = image::io::Reader::with_format(data, image::ImageFormat::Png);
 			reader.decode().map_err(|_| Error::ConversionFailure)
 		})?;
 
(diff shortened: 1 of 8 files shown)


complexspaces:
Hey there, thanks for the PR. Similar to the [first PR which tried this](https://github.com/1Password/arboard/pull/110#issuecomment-1694124434), can I ask that you do some basic compatibility testing to make sure this doesn't regress any major use cases? Some examples would be:
- Copying image to and from Firefox, Chrome, and Safari
- Copying image to and from Pages and/or Word
- Copying image to and from TextEdit

src/platform/osx.rs
```diff
@@ -295,14 +231,22 @@ impl<'clipboard> Set<'clipboard> {
...
-		let success = unsafe { self.clipboard.pasteboard.writeObjects(&image_array) };
+		let encoded = data.encode_as_png()?;
+		let success = unsafe {
+			let ns_data = {
+				NSData::initWithBytes_length(
+					NSData::alloc(),
+					encoded.as_ptr() as *mut c_void,
+					encoded.len(),
+				)
+			};
```
complexspaces, line 247: I believe it would be better to use [`NSData::from_vec`](https://docs.rs/objc2-foundation/latest/objc2_foundation/struct.NSData.html#method.from_vec) here because `block2` is already in use and this saves us a (possibly large) reallocation + memcpy.

complexspaces (changes requested):

src/platform/osx.rs
Gae24: Done

Gae24:
Unfortunately at the moment I don't have access to an Apple device.

complexspaces:
I'm marking this one with both `waiting on author` and `waiting on review` since in theory either of us can perform the testing. No pressure here though, I'm just doing some status triage and cleanup.

complexspaces:
I spent some time testing this today and my conclusion is that I believe we will need to continue to support TIFF on Apple platforms in order to retain compatibility with pasting images from Apple's own apps. 

In my testing I used Chrome, Firefox, Screenshot tool, and a bunch of Electron apps. All of these had no issues with providing data in a PNG format. However when I tried pasting an image supplied by `Pages.app`, it failed. After some debug printing I saw that it only offers it in TIFF or HEIC (which `image` doesn't support).

I did try a different approach (getting a bitmap of RGBA from `CGImage`) that would support either PNG or TIFF more transparently but ran into _another_ memory leak in Apple's image frameworks. When decoding TIFF data it leaks ~100 bytes, even inside of an autorelease pool. When decoding PNG there's no problems though. I guess it becomes a question of:
- How much do we care about consistency across platforms?
- How much do we want to drop the Rust `tiff` crate?
- Do people use TIFF enough in the "real world" via clipboard to run into the memory leaks?

<details>
<summary>Framework Memory Leak References</summary>

- https://web.archive.org/web/20140115104459/http://lists.apple.com/archives/cocoa-dev/2006/Nov/msg01416.html
- `https://github.com/Flipboard/FLAnimatedImage/issues/166`
- `https://github.com/vicc/chameleon/issues/75`
- https://stackoverflow.com/questions/62275191/memory-leak-after-call-cgcontextdrawimage
</details>

PR #167, opened 2024-11-16
Bump windows-sys to 0.59

Exotik850:
# Update windows-sys to 0.59

## Changes
- Upgrades windows-sys dependency from 0.48 to 0.59

## Breaking Changes
None. This is a semver-compatible update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,16 @@
 version = 3
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


complexspaces:
Hey there, thanks for the PR. I decided to take a bit more extensive approach here to allow for better `windows-sys` version compatibility going forward in https://github.com/1Password/arboard/pull/176. I added you as a co-author on that PR since I used some of your work here as a base and was inspired by it.

I'm going to close this as superseded, but versions of `arboard` going forward should be compatible with either `0.52` or `0.59` of `windows-sys` depending on what your lockfile currently uses. `0.48` has been dropped with the exception of `windows-targets 0.48` "lockfile text bloat" that will never get compiled.

Exotik850:
Of course! Thank you for adding me

PR #172, opened 2024-12-28
Bump all dependecies to newest version

eatradish:


complexspaces:
Hi there,

Generally I don't accept PRs that just mass update dependencies because it can be problematic for downstream crate user's dependency resolution or build times (or the local CI). In this specific case, the main dependencies being updated were:
- `windows-sys`: This was duplicated with #167 and I resolved it in #176 instead with more accompanying changes.
- `core-graphics`: This crate dependency was removed with #173 along with other updates and improvements to the macOS implementation.
- `wl-clipboard-rs`: I've gone ahead and updated it on its own in #177 instead of asking you for a whole new review cycle a month later for a one-line change.

I'm going to close this as superseded, but I appreciate the thought here. If I've missed something please file a followup issue.

PR #173, opened 2025-01-26
Update to objc2 v0.6

madsmtm:
This includes using the new crate `objc2-core-graphics`, which notably does not have the `CustomData` helper that `core-graphics`. This is probably for the better, as it allows us to avoid a double-boxing of the data.

This also:
- Bumps MSRV to 1.71 (required for `extern "C-unwind"` in `objc2`).
- Uses `dep:*` syntax (to avoid breaking changes from added/removed feature flags).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,11 @@ default = ["image-data"]
 image-data = [
     "dep:objc2-core-graphics",
     "dep:objc2-core-foundation",
-    "dep:image",
-    "dep:windows-sys",
+    "image",
+    "windows-sys",
+    "core-graphics",
 ]
-wayland-data-control = ["dep:wl-clipboard-rs"]
+wayland-data-control = ["wl-clipboard-rs"]
 
 # For backwards compat
 core-graphics = ["dep:objc2-core-graphics"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,19 +1,34 @@
 [package]
 name = "arboard"
 version = "3.4.1"
-authors = ["Artur Kovacs <kovacs.artur.barnabas@gmail.com>", "Avi Weinstock <aweinstock314@gmail.com>", "Arboard contributors"]
+authors = [
+    "Artur Kovacs <kovacs.artur.barnabas@gmail.com>",
+    "Avi Weinstock <aweinstock314@gmail.com>",
+    "Arboard contributors",
+]
 description = "Image and text handling for the OS clipboard."
 repository = "https://github.com/1Password/arboard"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 keywords = ["clipboard", "image"]
 edition = "2021"
-rust-version = "1.67.1"
+rust-version = "1.71.0"
 
 [features]
 default = ["image-data"]
-image-data = ["core-graphics", "image", "windows-sys"]
-wayland-data-control = ["wl-clipboard-rs"]
+image-data = [
+    "dep:objc2-core-graphics",
+    "dep:objc2-core-foundation",
+    "dep:image",
+    "dep:windows-sys",
+]
+wayland-data-control = ["dep:wl-clipboard-rs"]
+
+# For backwards compat
+core-graphics = ["dep:objc2-core-graphics"]
+windows-sys = ["dep:windows-sys"]
+image = ["dep:image"]
+wl-clipboard-rs = ["dep:wl-clipboard-rs"]
 
 [dependencies]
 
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -1,19 +1,34 @@
...
-wayland-data-control = ["wl-clipboard-rs"]
+image-data = [
+    "dep:objc2-core-graphics",
+    "dep:objc2-core-foundation",
+    "dep:image",
+    "dep:windows-sys",
+]
+wayland-data-control = ["dep:wl-clipboard-rs"]
+
+# For backwards compat
```
complexspaces, line 27: Thank you for thinking of this :)

src/platform/osx.rs
```diff
@@ -30,65 +30,67 @@ fn image_from_pixels(
...
+		ptr::{self, slice_from_raw_parts_mut, NonNull},
+	};
 
-	#[derive(Debug)]
-	struct PixelArray {
-		data: Vec<u8>,
+	unsafe extern "C-unwind" fn release(_info: *mut c_void, data: NonNull<c_void>, size: usize) {
+		let data = data.cast::<u8>();
+		let slice = slice_from_raw_parts_mut(data.as_ptr(), size);
+		// SAFETY: This is the same slice that we got from `Box::into_raw`.
```
complexspaces, line 50: Suggestion: You could use [`NonNull::slice_from_raw_parts`](https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html#method.slice_from_raw_parts) here to make it slightly more clear, and we'd still need the same number of `.cast()` and `.as_ptr()` calls.

src/platform/osx.rs
```diff
@@ -30,65 +30,67 @@ fn image_from_pixels(
...
-	impl CustomData for PixelArray {
-		unsafe fn ptr(&self) -> *const u8 {
-			self.data.as_ptr()
-		}
-		unsafe fn len(&self) -> usize {
-			self.data.len()
-		}
-	}
+	let provider = {
+		let pixels: Box<[u8]> = pixels.into();
```
complexspaces, line 55: Suggestion: [`Vec::into_boxed_slice`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.into_boxed_slice) doesn't require the explicit type annotation.

src/platform/osx.rs
```diff
@@ -30,65 +30,67 @@ fn image_from_pixels(
...
+		let len = pixels.len();
+		let pixels: *mut [u8] = Box::into_raw(pixels);
+		// Convert slice pointer to thin pointer.
+		let data_ptr = pixels.cast::<c_void>();
```
complexspaces, line 59: I'd like to double check my understanding of the soundness here: How is this clearing the metadata half of the fat pointer so that the system can't/won't read it? IIUC this is still passing a fat pointer to the system but the data half just happens to be at the start based on how slices are currently represented.

src/platform/osx.rs
```diff
@@ -216,7 +218,7 @@ impl<'clipboard> Get<'clipboard> {
 			let image_data = unsafe { self.clipboard.pasteboard.dataForType(NSPasteboardTypeTIFF) }
 				.ok_or(Error::ContentNotAvailable)?;
 
-			let data = Cursor::new(image_data.bytes());
+			let data = Cursor::new(image_data.to_vec());
```
complexspaces, line 221: Question: Couldn't this be [`as_bytes_unchecked`](https://docs.rs/objc2-foundation/0.3.0/objc2_foundation/struct.NSData.html#method.as_bytes_unchecked) to avoid the (possibly quite large) allocation only for it to be dropped right after decoding?

My understanding is that the system is not going to ever modify the data; its been copied from the pasteboard server into our process.

complexspaces (commented):
Thank you for work again! I like how this looks for the most part but had a few questions around the modified `unsafe`.

src/platform/osx.rs
madsmtm: The best way of writing this would be [`<*const [T]>::as_ptr`](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ptr), but that's unstable.

I'm convinced that a slice pointer ->  pointer cast is safe and does this, and the linked (unstable) docs above do also document this ("This is equivalent to casting `self` to `*const T`, but more type-safe."), but I looked at the reference and couldn't actually find any notes on this?

madsmtm: Nvmd, the docs are here:
https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#r-expr.as.pointer.discard-metadata

src/platform/osx.rs
madsmtm: Yeah, it'd be safe to use `as_bytes_unchecked`, have changed it to that (I just thought I'd start with the definitely safe option).

src/platform/osx.rs
complexspaces: Thanks for digging that up. I agree then this is correct and sound 👍

complexspaces (approved):
Looks good now, thanks!

With this update I should also be able to take another stab at iOS support, hopefully soon.

torokati44:
I would appreciate a release on crates.io with this bump, at your earliest convenience. Thank you! :heart_hands:

torokati44:
Release please! :pleading_face: :pray:

complexspaces:
Released in [v3.5.0](https://crates.io/crates/arboard/3.5.0) on crates.io. Sorry again for the delay.

PR #176, opened 2025-02-13
Update windows-sys version compatibility

complexspaces:
This PR technically updates `windows-sys`, but more importantly it makes some changes to allow `arboard` to work with _either_ `0.52` (more common) or `0.59` versions of the crate. This change's goal is to make it easier to balance the desires of parties who want the latest version in their tree (and to help the ecosystem move along) with those who are stuck with older versions and don't want their build times or lockfiles damaged.

It accomplishes this via three parts:
- Making changes to the `windows.rs` module to abstract over some function parameter and return value changes that have been made in the past.
- Using a range specifier in `arboard`'s `Cargo.toml`
- Implementing CI testing to prevent accidentally using newer features that don't exist in `0.52` which would break downstream builds. This is the most important part as it helps verify our code is actually compatible with this version range.

The idea is based off many similar changes being made across crates.io, including [winapi-util](https://github.com/BurntSushi/winapi-util/pull/19),  [errno](https://github.com/lambda-fairy/rust-errno/pull/98), [is-terminal](https://github.com/sunfishcode/is-terminal/pull/42/), and many others to reduce version churn for users.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,12 @@ jobs:
       - name: Run `cargo clippy` with all features
         run: cargo clippy --verbose --all-features -- -D warnings -D clippy::dbg_macro
 
+      - name: Run `cargo clippy` with dependency version checks
+        if: ${{ matrix.rust_version == 'stable' }}
+        run: |
+          cargo update -p windows-sys
+          cargo clippy --verbose --all-features -- -D warnings -D clippy::dbg_macro 
+
   test:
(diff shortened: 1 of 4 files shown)


complexspaces:
Looking at the CI run you can see that we download + check `0.59` of `windows-sys` in the new job, which succeeds, confirming there are no cargo version conflicts and that `arboard`'s source is compatible with both library versions:
image

Locally running the same steps demonstrates the same.

PR #179, opened 2025-02-25
Add `file_list` to `Get` interface

### LINKED ISSUE #51: Files & directories support by path
I'm interested in getting the paths of files & directories copied from explorer/finder/etc. Is this a use case that will be supported?

I just tried on MacOS, and through the `get_text()` API I got the **names** of the files/dirs separated by `\r`, but not the full paths, nor a way to distinguish between regular text and file paths.

Thanks for creating and maintaining this library!

Gae24:
Opening as a draft since osx and wayland are still to do.
While Windows would simply return file paths, X11 returns a uri-list, should we maybe decide a single format to use in all platforms?
Close #51

### DIFF
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -38,6 +38,10 @@ fn encode_as_png(image: &ImageData) -> Result<Vec<u8>, Error> {
 	Ok(png_bytes)
 }
 
+fn extract_paths_from_uri_list(uri_list: String) -> Vec<PathBuf> {
+	uri_list.lines().filter_map(|s| s.strip_prefix("file://").map(PathBuf::from)).collect()
+}
+
 /// Clipboard selection
 ///
 /// Linux has a concept of clipboard "selections" which tend to be used in different contexts. This
@@ -132,7 +136,11 @@ impl<'clipboard> Get<'clipboard> {
 	}
 
 	pub(crate) fn file_list(self) -> Result<Vec<PathBuf>, Error> {
-		todo!()
+		match self.clipboard {
+			Clipboard::X11(clipboard) => clipboard.get_file_list(self.selection),
+			#[cfg(feature = "wayland-data-control")]
+			Clipboard::WlDataControl(clipboard) => clipboard.get_file_list(self.selection),
+		}
 	}
 }
 
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -2,6 +2,7 @@ use std::{borrow::Cow, path::PathBuf, time::Instant};
 
 #[cfg(feature = "wayland-data-control")]
 use log::{trace, warn};
+use percent_encoding::percent_decode_str;
 
 #[cfg(feature = "image-data")]
 use crate::ImageData;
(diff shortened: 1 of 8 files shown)


complexspaces:
I think it would probably be best if we returned `PathBuf` in these APIs, so it is possible to normalize the Linux contents across the various implementations (ie do they use a standard URI scheme)? On macOS `NSUrl` lets you extract a path out of it.

As a secondary question: Do I understand that the primary use case for this is to receive a list of file paths that have been drag-and-dropped into an app's window? So for macOS we'd be looking for `NSPasteboardTypeFileURL`?

Gae24:
Not just for drag and drop, e.g. when trying to copy a file from file explorer, this is the data placed in the clipboard.

complexspaces:
Ah right, thanks for clarifying 👍.

Gae24:
I can't test osx code, so I would like to know if like on Linux file paths are prefixed with `file://`.

complexspaces:
I tested the OSX code this morning using a modified example and confirmed it returns file paths, with no prefixes:
```
[
"/Applications/Keynote.app", 
"/System/Applications/Maps.app", 
"/System/Applications/Messages.app",
"/System/Applications/Notes.app", 
"/Applications/Pages.app"
]
```

This is because you're using the `.path()` method on `NSUrl` which is doing what we want.

Gae24:
Thank you for testing

src/platform/linux/mod.rs
```diff
@@ -38,6 +38,10 @@ fn encode_as_png(image: &ImageData) -> Result<Vec<u8>, Error> {
 	Ok(png_bytes)
 }
 
+fn extract_paths_from_uri_list(uri_list: String) -> Vec<PathBuf> {
```
complexspaces, line 41: Would you mind adding a unit test for this function?

src/platform/osx.rs
```diff
@@ -237,6 +241,34 @@ impl<'clipboard> Get<'clipboard> {
...
+		autoreleasepool(|_| {
+			let class_array = NSArray::from_slice(&[NSURL::class()]);
+			let options = NSDictionary::from_slices(
+				&[unsafe { NSPasteboardURLReadingFileURLsOnlyKey }],
+				&[NSNumber::new_bool(true).as_ref()],
+			);
+			let objects = unsafe {
+				self.clipboard
+					.pasteboard
+					.readObjectsForClasses_options(&class_array, Some(&options))
```
complexspaces, line 255: There seems to be a behavior difference between the current revision of changes and the last set. Previously if I didn't have a set of files copied, `file_list` returned an error on macOS. However it now always returns `Ok([])` which isn't what we want since I think people would expect to see `ContentNotAvailable` in those cases.

Gae24:
Upon further testing on Linux, since file paths are exposed as `text/uri-list` they are percent encoded.
I think we should decode them, we can use [percent-encoding](https://crates.io/crates/percent-encoding).
Does macOS  has the same behaviour?

src/platform/osx.rs
Gae24: Done

complexspaces:
> I think we should decode them, we can use [percent-encoding](https://crates.io/crates/percent-encoding).

That sounds good to me.

>  Does macOS has the same behaviour?

No, the `path` function handles converting these to normal characters instead of URL encoded ones. For example this is noted in [Apple's documentation](https://developer.apple.com/documentation/foundation/nsurl/1408809-path?language=objc):
> This property contains the path, unescaped using the [stringByReplacingPercentEscapesUsingEncoding:](https://developer.apple.com/documentation/foundation/nsstring/1407783-stringbyreplacingpercentescapesu?language=objc) method. If the receiver does not conform to RFC 1808, this property contains nil.

complexspaces (approved):
Sorry for the delay, this looks good now.

Thanks for the work!

complexspaces:
Released in [v3.5.0](https://crates.io/crates/arboard/3.5.0) on crates.io.

PR #180, opened 2025-03-09
use `Error::unknown` on Linux and macOS code too

Gae24:
Unsure why this method wasn't used more, just code cleanup.

### DIFF
--- a/src/platform/linux/wayland.rs
+++ b/src/platform/linux/wayland.rs
@@ -72,7 +72,7 @@ impl Clipboard {
 
 			Err(PasteError::PrimarySelectionUnsupported) => Err(Error::ClipboardNotSupported),
 
-			Err(err) => Err(Error::Unknown { description: err.to_string() }),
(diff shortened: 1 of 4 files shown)


src/platform/linux/wayland.rs
```diff
@@ -72,7 +72,7 @@ impl Clipboard {
 
 			Err(PasteError::PrimarySelectionUnsupported) => Err(Error::ClipboardNotSupported),
 
-			Err(err) => Err(Error::Unknown { description: err.to_string() }),
+			Err(err) => Err(Error::unknown(err.to_string())),
```
complexspaces, line 75: Suggestion here and elsewhere: `Error::unknown` allows anything that can be turned into a string so you should be able to cleanup the noise further and remove all the `.to_string()` calls.

complexspaces (changes requested):

complexspaces (approved):
Thank you for cleaning this up 🎉

PR #181, opened 2025-03-19
add `file_list` to `Set` interface

Gae24:
We should decide what type of input we accept.
On windows `clipboard_win::raw::set_file_list` is used which has as parameter `&[impl AsRef<str>]` and doesn't check if the input is actually a path; on linux and osx the parameter is of type `&[impl AsRef<Path>]`  and `canonicalize()` is called so non existing paths get discarded

### DIFF
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -1,8 +1,12 @@
-use std::{borrow::Cow, path::PathBuf, time::Instant};
+use std::{
+	borrow::Cow,
+	path::{Path, PathBuf},
+	time::Instant,
+};
 
 #[cfg(feature = "wayland-data-control")]
 use log::{trace, warn};
-use percent_encoding::percent_decode;
+use percent_encoding::{percent_decode, utf8_percent_encode, AsciiSet, CONTROLS};
 
 #[cfg(feature = "image-data")]
 use crate::ImageData;
@@ -50,6 +54,39 @@ fn paths_from_uri_list(uri_list: Vec<u8>) -> Vec<PathBuf> {
 		.filter_map(|s| percent_decode(s).decode_utf8().ok())
 		.map(|decoded| PathBuf::from(decoded.as_ref()))
 		.collect()
+}
+
+fn paths_to_uri_list(file_list: &[impl AsRef<Path>]) -> Result<String, Error> {
+	// The characters that require encoding, which includes £ and € but they can't be added to the set.
+	const ASCII_SET: &AsciiSet = &CONTROLS
+		.add(b'#')
+		.add(b';')
+		.add(b'?')
+		.add(b'[')
+		.add(b']')
+		.add(b' ')
+		.add(b'\"')
+		.add(b'%')
+		.add(b'<')
+		.add(b'>')
+		.add(b'\\')
+		.add(b'^')
+		.add(b'`')
+		.add(b'{')
+		.add(b'|')
+		.add(b'}');
+
+	file_list
+		.iter()
+		.filter_map(|path| {
+			path.as_ref().canonicalize().ok().and_then(|abs_path| {
+				abs_path
+					.to_str()
+					.map(|str| format!("file://{}", utf8_percent_encode(str, ASCII_SET)))
+			})
+		})
+		.reduce(|uri_list, uri| uri_list + "\n" + &uri)
+		.ok_or(Error::ConversionFailure)
 }
 
 /// Clipboard selection
@@ -237,6 +274,19 @@ impl<'clipboard> Set<'clipboard> {
 			#[cfg(feature = "wayland-data-control")]
 			Clipboard::WlDataControl(clipboard) => {
 				clipboard.set_image(image, self.selection, self.wait, self.exclude_from_history)
+			}
+		}
+	}
+
+	pub(crate) fn file_list(self, file_list: &[impl AsRef<Path>]) -> Result<(), Error> {
+		match self.clipboard {
+			Clipboard::X11(clipboard) => {
+				clipboard.set_file_list(file_list, self.selection, self.wait)
+			}
+
+			#[cfg(feature = "wayland-data-control")]
+			Clipboard::WlDataControl(clipboard) => {
+				clipboard.set_file_list(file_list, self.selection, self.wait)
 			}
 		}
 	}
--- a/src/platform/linux/mod.rs
+++ b/src/platform/linux/mod.rs
@@ -1,12 +1,13 @@
 use std::{
 	borrow::Cow,
+	os::unix::ffi::OsStrExt,
 	path::{Path, PathBuf},
 	time::Instant,
 };
 
 #[cfg(feature = "wayland-data-control")]
 use log::{trace, warn};
-use percent_encoding::{percent_decode, utf8_percent_encode, AsciiSet, CONTROLS};
+use percent_encoding::{percent_decode, percent_encode, AsciiSet, CONTROLS};
 
 #[cfg(feature = "image-data")]
 use crate::ImageData;
@@ -79,10 +80,8 @@ fn paths_to_uri_list(file_list: &[impl AsRef<Path>]) -> Result<String, Error> {
 	file_list
 		.iter()
 		.filter_map(|path| {
-			path.as_ref().canonicalize().ok().and_then(|abs_path| {
-				abs_path
-					.to_str()
-					.map(|str| format!("file://{}", utf8_percent_encode(str, ASCII_SET)))
+			path.as_ref().canonicalize().ok().map(|path| {
+				format!("file://{}", percent_encode(path.as_os_str().as_bytes(), ASCII_SET))
 			})
 		})
 		.reduce(|uri_list, uri| uri_list + "\n" + &uri)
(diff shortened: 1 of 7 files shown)


complexspaces:
It seems like the best generic interface we could expose is `impl Iterator<Item = impl AsRef<Path>>]` because its the most typesafe, but that would require a re-collection (`.collect<Vec<_>>`) on Windows because of the interface `clipboard-win` gives us (despite internally just looping over the slice). I think its advantages are:
- Allows callers with well-typed libraries/crates using `Path`/`PathBuf` no friction/extra unwraps.
- Allows callers with borrowed data in their collections (`&str`) no friction.
- Allows callers with owned data in their collections (`String`) no friction.
- Allows `arboard` on non-Windows platforms to skip a collection allocation.

I was experimenting with the above in [this playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b7d4a301c6c2762e9834a395fbd6db0c), for reference. Do you think its worth just dealing with the current limitation on Windows and seeing if it can be removed via either a PR to `clipboard-win` or re-implementing similar in `arboard` so that we can have a flexible interface for the future?

Gae24:
It depends on which behavior we want. On Linux and MacOs the argument is `&[impl AsRef<Path>]` due to the use of `canonicalize()` since I thought we shouldn't write invalid paths.
If we want align Windows to the same behaviour, I say let's implement our own version.

complexspaces:
That sounds reasonable to me; the Win32 code for this isn't super bad. We already have some primitives for clipboard memory allocation due to image support.

One thought came to me over the weekend I'd like to ask about regarding invalid paths: Can you think of any use cases or needs for "dropping" a list of files on the clipboard that aren't proper file URIs, such as app-specific URIs (`onepassword://`, for example). It might be out of scope for this addition too, which is fine.

Gae24:
Certainly, we can expand support for more uri types later on.

Gae24:
Had to modify a bit `Cargo.toml` and had to move out of `image_data` a couple of functions.

For some reason, `clipboard_win` doesn't place the data in a STGMEDIUM, should we do it?

complexspaces:
It doesn't look like Chromium does, so I don't think we need to. They create a `STDMEDIUM` structure [here](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_util_win.cc;l=610-654?q=CF_HDROP&ss=chromium%2Fchromium%2Fsrc) but then when actually writing the data [only the `hGlobal`](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_win.cc;l=798-802;drc=10d39cde3307a7249d2dd106684cf5467692f8de;bpv=0;bpt=1) part is actually used.

complexspaces:
Sorry, missed your comment. I learn towards not wanting to introduce UNC paths because since we're very reliant on other applications reading the pasteboard to know how to handle them correctly.

src/platform/linux/mod.rs
```diff
@@ -52,6 +56,39 @@ fn paths_from_uri_list(uri_list: String) -> Vec<PathBuf> {
...
+		.add(b'{')
+		.add(b'|')
+		.add(b'}');
+
+	file_list
+		.iter()
+		.filter_map(|path| {
+			path.as_ref().canonicalize().ok().and_then(|abs_path| {
+				abs_path
+					.to_str()
```
complexspaces: Suggestion: We probably want to use `PathExt` to get the bytes of the path instead of the string form for better compatibility. `percent_encoding::percent_encode` should handle that.

src/platform/linux/wayland.rs
```diff
@@ -241,4 +245,23 @@ impl Clipboard {
...
 	}
+
+	pub(crate) fn set_file_list(
+		&self,
+		file_list: &[impl AsRef<Path>],
+		selection: LinuxClipboardKind,
+		wait: WaitConfig,
+	) -> Result<(), Error> {
+		let files = paths_to_uri_list(file_list)?;
+		let uri_mime = MimeType::Specific(String::from("text/uri-list"));
```
complexspaces: Suggestion: Let's refactor `text/uri-list` out into a constant now.

src/platform/linux/x11.rs
```diff
@@ -1063,6 +1063,19 @@ impl Clipboard {
...
+		&self,
+		file_list: &[impl AsRef<Path>],
+		selection: LinuxClipboardKind,
+		wait: WaitConfig,
+	) -> Result<()> {
+		let files = paths_to_uri_list(file_list)?;
+
+		let data =
+			vec![ClipboardData { bytes: files.into_bytes(), format: self.inner.atoms.URI_LIST }];
+		self.inner.write(data, selection, wait)
```
complexspaces: With the Linux clipboard exclusion support merged since this was opened, this will need refactored to look more like `self.set_text` where if exclusions are told to be set, it pushes that MIME set into the `data` `Vec` too.

src/platform/linux/wayland.rs
```diff
@@ -241,4 +245,23 @@ impl Clipboard {
...
+		let mut opts = Options::new();
+		opts.foreground(matches!(wait, WaitConfig::Forever));
+		opts.clipboard(selection.try_into()?);
+		let source = Source::Bytes(files.into_bytes().into_boxed_slice());
```
complexspaces: This also needs to consider clipboard exclusions now.

src/platform/windows.rs
```diff
@@ -487,6 +451,55 @@ mod image_data {
...
+fn last_error(message: &str) -> Error {
+	let os_error = io::Error::last_os_error();
+	Error::unknown(format!("{message}: {os_error}"))
+}
+
+/// An abstraction trait over the different ways a Win32 function may return
+/// a value with a failure marker.
+///
+/// This is primarily to abstract over changes in `windows-sys` versions and unify how
+/// error handling is done in the above image code.
```
complexspaces: Nit: This comment is now outdated with the trait being moved.

src/platform/windows.rs
```diff
@@ -702,6 +715,69 @@ impl<'clipboard> Set<'clipboard> {
...
+		if paths.is_empty() {
+			return Err(Error::ConversionFailure);
+		}
+
+		// Add space for the final null character
+		data_len += std::mem::size_of::<u16>();
+
+		unsafe {
+			let h_global = global_alloc(data_len)?;
+			let data_ptr = global_lock(h_global)?;
```
complexspaces, line 757: Question: We need to `GlobalUnlock` the data after we finish writing to it, right?

src/platform/windows.rs
```diff
@@ -702,6 +715,69 @@ impl<'clipboard> Set<'clipboard> {
...
+
+			(data_ptr as *mut DROPFILES).write(dropfiles);
+
+			let mut ptr = data_ptr.add(DROPFILES_HEADER_SIZE) as *mut u16;
+
+			for wide_path in paths {
+				for wchar in wide_path {
+					ptr.write(wchar);
+					ptr = ptr.add(1);
+				}
```
complexspaces: Suggestion: Let's use `ptr::copy_nonoverlapping` here like in the image code instead to avoid manually writing out `memcpy`.

src/platform/windows.rs
```diff
@@ -826,3 +902,101 @@ fn wrap_html(ctn: &str) -> String {
...
+			let mut wide = Vec::with_capacity(buf.len() + 1);
+			wide.extend_from_slice(buf);
+			wide.push(0);
+
+			let hr = unsafe { PathCchStripPrefix(wide.as_mut_ptr(), wide.len()) };
+			// On success truncate invalid data
+			if hr == S_OK {
+				if let Some(end) = wide.iter().position(|c| *c == 0) {
+					// Retain NULL character
+					wide.truncate(end + 1)
```
complexspaces, line 936: Question: Its not clear to me how this removes the now invalid now? IIUC it wouldn't touch any data at the start.

src/platform/windows.rs
```diff
@@ -826,3 +902,101 @@ fn wrap_html(ctn: &str) -> String {
...
+				}
+			}
+			wide
+		},
+	)
+}
+
+fn fill_utf16_buf<F1, F2, T>(mut f1: F1, f2: F2) -> Option<T>
```
complexspaces, line 945: Please leave a comment with a Git permalink to the stdlib path + commit this was vendored from.

complexspaces (changes requested):

src/platform/windows.rs
Gae24: On success, the resulting path will be shorter than the length of the input buffer, so without the null check, garbage data will be included.
I suppose internally, it does something like memmove/shifting the contents of the buffer.

src/platform/linux/x11.rs
Gae24: Do I need to do it on Windows, too?

complexspaces: Oh good catch, yes it does. Sorry for not mentioning that originally.

src/platform/windows.rs
complexspaces: That sounds reasonable. If this has the correct behavior I'm not that worried about it in the end.

complexspaces:
The new set of changes look good to me. I'm happy to approve this and merge once the last thread is resolved.

complexspaces (approved):

complexspaces:

Thanks again for the new feature! I hope `servo` continues to make good progress :)

PR #182, opened 2025-03-27
Allow Compilation on Android Termux via Conditional Compilation Fix

RyderFreeman4Logos:
This pull request updates the conditional compilation settings in Cargo.toml, src/lib.rs, and src/platform/mod.rs. By removing Android from the exclusion list, the arboard library (and dependent applications such as sendme) can now compile successfully in Termux environments on Android.

Please note that this change is intended solely to enable the software to compile on Termux. If full clipboard functionality is required on Termux, it can be implemented relatively easily since Termux provides its own clipboard API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,7 +81,7 @@ image = { version = "0.25", optional = true, default-features = false, features 
     "tiff",
 ] }
 
-[target.'cfg(all(unix, not(any(target_os="macos", target_os="android", target_os="emscripten"))))'.dependencies]
+[target.'cfg(all(unix, not(any(target_os="macos", target_os="emscripten"))))'.dependencies]
 log = "0.4"
 x11rb = { version = "0.13" }
 wl-clipboard-rs = { version = "0.9.0", optional = true }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,7 +20,7 @@ mod platform;
 
 #[cfg(all(
 	unix,
-	not(any(target_os = "macos", target_os = "android", target_os = "emscripten")),
+	not(any(target_os = "macos", target_os = "emscripten")),
 ))]
 pub use platform::{ClearExtLinux, GetExtLinux, LinuxClipboardKind, SetExtLinux};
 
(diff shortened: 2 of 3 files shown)


RyderFreeman4Logos:
> Hi there,
> 
> Do you happen to know if the native Android `ClipboardManager` APIs (Java/Kotlin) work when running inside a Termux environment on Android, or do you need to use Termux's specific ones?

Hi there,

That's a great question regarding the clipboard implementation on Termux.

Based on my understanding, programs running within the standard Termux environment generally cannot directly invoke the native Android `ClipboardManager` APIs (Java/Kotlin). This necessitates using the Termux-specific mechanisms, such as the `termux-clipboard-get` and `termux-clipboard-set` commands, which I alluded to in the PR description.

Furthermore, it's important to note that even for these Termux commands (`termux-clipboard-get`/`set`) to function, the end-user must have the separate `termux-api` apk installed (available here: https://github.com/termux/termux-api) and grant it the appropriate permissions via the Android system. So, while using the Termux APIs is the path forward for clipboard interaction within this environment, it does come with this external dependency and user setup requirement.

Just to reiterate, my current PR focuses only on enabling *compilation* within Termux. The actual implementation of clipboard functionality using these Termux APIs (and handling the `termux-api` dependency) would be a separate, subsequent step.

Hope this information is helpful! Let me know if you have any other questions.

complexspaces:
Sorry for my delay, thanks for the additional information and context since I was very unfamiliar with Termux.

Please see https://github.com/1Password/arboard/issues/191 which describes the pre-requesties for Termux support getting merged, similar to #171. As mentioned there, my intent is to have pending platform backend additions all wrapped up no later then end of summer.

> Hope this information is helpful! Let me know if you have any other questions.

I think you answered everything, thanks!

> Just to reiterate, my current PR focuses only on enabling compilation within Termux. The actual implementation of clipboard functionality using these Termux APIs (and handling the termux-api dependency) would be a separate, subsequent step.

Under my current plan of making Termux an `arboard` tier 3 platform, this would be perfectly since it still compiles.

With that said, if the PRs will be split up, my strong preference would be to introduce the Cargo feature flagging required, or some other reliable mechanism for Termux detection, in this PR so its clear ahead of time how this backend will interact with the native app-hosted implementation. It may be good to draw some inspiration from how the Wayland and X11 split is done on Linux today.

PR #186, opened 2025-06-07
linux/x11: Don't stop worker thread if handling selection request fails

crumblingstatue:
Just because one request fails, that doesn't mean we can't handle subsequent requests.
Stopping the worker thread on failure means that every subsequent request will fail, which is usually undesirable.

# Unresolved questions

I don't know if this has any undesirable side effects, since I'm not an X11/clipboard expert.
This requires careful review.

- [ ] Should we have some kind of way to programmatically get the error, rather than logging it?

### DIFF
--- a/src/platform/linux/x11.rs
+++ b/src/platform/linux/x11.rs
@@ -764,7 +764,10 @@ fn serve_requests(context: Arc<Inner>) -> Result<(), Box<dyn std::error::Error>>
 					context.atom_name_dbg(event.target),
 				);
 				// Someone is requesting the clipboard content from us.
-				context.handle_selection_request(event).map_err(into_unknown)?;
+				if let Err(e) = context.handle_selection_request(event).map_err(into_unknown) {
+					error!("Failed to handle selection request: {e}");
+					continue;
+				}
 
 				// if we are in the progress of saving to the clipboard manager
 				// make sure we save that we have finished writing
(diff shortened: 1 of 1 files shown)


crumblingstatue:
> Out of curiosity do you have any additional context or examples of where this helps in the wild?

I have an [image organizer app](<https://github.com/crumblingstatue/cowbump/>), and it allows copying images to the clipboard.
However, if a very large image gets copied, it gets a `Maximum request length exceeded` error, which kills the worker thread.

With this patch, it won't kill the worker thread, so the user can keep copying other images (and strings, etc.).

crumblingstatue:
> it seems like most of the enum variants are fatal? I'm not an X11 expert but my reading is that anything prefixed with ` XCB_CONN_` means future requests won't work again because the server has closed our socket. Does that match up with your understanding?

Well, x11rb is its own thing, not bindings to XCB, so I wouldn't take the XCB documentation as "upstream".

complexspaces (approved):
Thanks again!

Before I merge this, do you mind confirming that you did actually test that future copy operations work after the length exceeded error is logged? After that we're good. I couldn't exactly tell if you had live tested it or were just describing how it _would_ behave.

crumblingstatue:
If there is a way to check that the connection is closed, we could add a check for that in the event loop, and terminate if the connection is closed. 
I'll investigate when I get home.

complexspaces (approved):

complexspaces:
This behavior change is now present in [3.6.0](https://crates.io/crates/arboard/3.6.0) of `arboard` on crates.io.

PR #189, opened 2025-06-16
Android support

jb55:
This is a rebase and rework of:

- #171 

I had to update a few dependencies so that jni's window-sys wouldn't conflict, but after I did that the CI seems to pass.

I also fixed an issue where paste wasn't working in the original PR

Here's a video of it working in my egui android app:

https://github.com/user-attachments/assets/c83b6ae6-406b-4476-8756-dfda4293c7c5

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -89,6 +89,7 @@ impl std::fmt::Debug for Error {
 	}
 }
 
+#[cfg(not(target_os = "android"))]
 impl Error {
 	pub(crate) fn unknown<M: Into<String>>(message: M) -> Self {
 		Error::Unknown { description: message.into() }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -91,6 +91,11 @@ image = { version = "0.25", optional = true, default-features = false, features
 parking_lot = "0.12"
 percent-encoding = "2.3.1"
 
+[target.'cfg(target_os = "android")'.dependencies]
+ndk-context = "0.1.1"
+# latest crate-compatible version of jni is not on crates.io, update this when it is
+jni = { git = "https://github.com/jni-rs/jni-rs", rev = "1a4a345b1e65aca405e856046ff2661d37edde6d" }
```
complexspaces, line 97: I'm not entirely sure what to do about this off the top of my head. When you say "crate compatible" do you mean that Cargo won't build Windows or Android when `jni` brings in `windows-sys` 0.48 because of the range restriction earlier in this file?

Off the top of my head, I wonder if its feasible/possible to expand the version range here to support `0.48` too with more internal resolution workarounds...

jb55: I tried with the version on crates.io and I was getting tons of windows-sys conflicting errors. switching to master on jni fixed it. 

I found this related issue:

- https://github.com/jni-rs/jni-rs/issues/523

PR #197, opened 2025-08-01
Linux: avoid converting uri-list to a String

Gae24:
Not really needed, in addition if some of the paths contained non utf8 data we returned nothing.


complexspaces (approved):
This is a nice set of cleanup, thank you!

complexspaces:

Thanks again for the contribution, I marked it as a bugfix in the release notes since non-UTF8 paths are a completely valid use case to support.

Issue #122, opened 2023-11-22
Retrieving an image from the clipboard and saving it results in pixel distortion


zibo-chen:
![image](https://github.com/1Password/arboard/assets/58510061/a0fb1b74-4b1c-4f94-846d-cb6ee4fe9f8b)

After retrieving the image from the clipboard, I loaded it with the image library and saved it as a PNG, and there are misplaced pixels in the bottom left corner.

zibo-chen:
The pixels on the right side of the image have also moved to the left.

complexspaces:
Hi, thanks for opening this issue. 

Can I ask what OS (and desktop environment, if applicable) you are seeing this issue in?

zibo-chen:
This issue can be reproduced on both Windows 10 and Windows 11.

Klavionik:
I think I'm facing the same issue with arboard 3.3.1. This happens on Windows 10 if I place an image into the clipboard using a bit of C#; I believe this doesn't happen if I use, let's say, Copy image in Chrome. 

I noticed one curious thing: if a distorted image has 1px height (no matter the width), it will always start as `[255, 0, 0, 0, 0, 255, 0, 0, 0, 0, 255, 0]`. This very 12 bytes. If, however, the image has a shape of square/rectangle/whatever, the distorted bytes are going to be somewhere in the middle.

This is how bytes look for a 14x1 red line (`<byte number>. <pasted byte> -- <original byte> <match_or_not>`):

![14x1_red_line](https://github.com/1Password/arboard/assets/56698047/96799c67-dcd6-42b2-9daf-6bf03327e4c3)

And this is how it looks for a 10x10 red square:

![10x10_red_square](https://github.com/1Password/arboard/assets/56698047/e0d01606-1ae1-4172-9967-2a1ffec0ee4d)

zibo-chen:
I have found a reliable way to reproduce the pixel distortion issue when saving an image from the clipboard.

Steps to reproduce:

Ensure that Hyper-V is enabled on your Windows 10 or Windows 11 system.
Connect to a virtual machine using Hyper-V.
Use the built-in snipping tool of Windows 10 to take a screenshot.
Copy the screenshot to the clipboard.
Following these steps, the bug can be consistently reproduced both on Windows 10 and Windows 11 environments. It seems that the issue might be related to how the snipping tool interacts with the clipboard when Hyper-V is active, or how the clipboard data is being handled by the system in this specific virtualized context.

[...]

PR #198, opened 2025-08-09
Windows: fix parsing of 24-bit bitmaps and prefer PNG when available

### LINKED ISSUE #192: Consider switching to retrieving PNG on Windows
Hi! Thanks for your work maintaining this crate. 

I was wondering if there's any reason not to fetch image data in PNG format on Windows. For context, I've run into a case where images copied from Firefox lose transparency data, I think due to a longstanding bug ([this one](https://bugzilla.mozilla.org/show_bug.cgi?id=460969)? 17 years old!!); the results of `get().image()` have all zeroes in the alpha channel. My sense from a quick google is that implementation of transparency for CF_DIBV5 is inconsistent at best, and switching to PNG would make things more robust. But I know next to nothing about Windows 😁    

Related: I saw that https://github.com/1Password/arboard/pull/141/files has already implemented _setting_ images in PNG.

If this seems like a reasonable idea to you, I'd be happy to have a go at the implementation!

wcassels:
Fixes #192

I tested on commits before 74ece6a that copying from Chrome/snipping tool/Discord/a few more still work, and that copying from Paint/Firefox works too (with transparent backgrounds filled in with black).

I then tested on the PNG-supporting commit that everything still works, with additional support for transparency from Firefox

Aside: this doesn't feel like the perfect solution. Hypothetically, we could see a bitmap with bitCount = 32, compression = BI_RGB, non-zero alpha mask, and nonsense/zero bytes in the alpha channel that upstream (who's following the spec strictly) wants us to ignore; I think this PR breaks support for cases like this (when PNG's unavailable). But I think supporting copying from firefox/paint makes this a clear improvement, and the tests I added make it easy to iterate safely if/when the next person runs into these issues :)


src/platform/windows.rs
```diff
@@ -186,138 +188,81 @@ mod image_data {
...
+	// https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapv5header
+	// According to the docs, when bV5Compression is BI_RGB, "the high byte in each DWORD
+	// is not used".
+	// This seems to not be respected in the real world. For example, Chrome, and Chromium
+	// & Electron-based programs send us BI_RGB headers, but with bitCount=32 - and important
+	// transparency bytes in the alpha channel.
+	//
+	// Apparently, it's our job as the consumer to do the right thing. This method fiddles
+	// with the header a bit in these cases, then `image` handles the rest.
+	fn maybe_tweak_header(dibv5: &mut [u8]) {
```
complexspaces, line 178: Question: Is there a reason we don't cast this to a mutable header structure and are instead doing manual offsets?

src/platform/windows.rs
```diff
@@ -186,138 +188,81 @@ mod image_data {
...
-			DIB_RGB_COLORS,
-		);
-		if lines == 0 {
-			Err(Error::unknown("Could not get the bitmap bits, GetDIBits returned 0"))
-		} else {
-			Ok(lines)
-		}
+	pub(super) fn read_png(data: &[u8]) -> Result<ImageData<'static>, Error> {
+		let decoder = PngDecoder::new(std::io::Cursor::new(data))
+			.map_err(|_| Error::unknown("Failed to read PNG header"))?;
```
complexspaces: Suggestion: This should be a conversion error.

src/platform/windows.rs
```diff
@@ -186,138 +188,81 @@ mod image_data {
...
+		let (width, height) = decoder.dimensions();
+
+		let bytes = DynamicImage::from_decoder(decoder)
+			.map_err(|_| Error::unknown("Failed to decode PNG"))?
```
complexspaces: Same here.

src/platform/windows.rs
```diff
@@ -186,138 +188,81 @@ mod image_data {
...
-		} else {
-			Ok(hbitmap)
+		let header_size = size_of::<BITMAPV5HEADER>();
+		if dibv5.len() < header_size {
+			return Err(Error::unknown("When reading the DIBV5 data, it contained fewer bytes than the BITMAPV5HEADER size. This is invalid."));
 		}
+		maybe_tweak_header(dibv5);
+
+		let decoder = BmpDecoder::new_without_file_header(std::io::Cursor::new(&*dibv5))
+			.map_err(|_| Error::unknown("Failed to read bitmap header"))?;
```
complexspaces: Let's use `ConversionError` in these places too.

complexspaces (changes requested):

src/platform/windows.rs
wcassels: Not really - guess it's a question of style. Agree interacting directly with the header's fields would be cleaner and easier to read but obviously introduces unsafe code

Would you prefer a direct cast to a &mut BITMAPV5HEADER? I see that the existing code does this casting (to an immutable ref), but doesn't check alignment. Do you know why that's ok?

complexspaces: I personally prefer to work with data in terms of structured fields, but the alignment check is a good callout. I think it may have been historically assuming that the system only gives us well-formed data but I don't know if that's good enough to rely on. To be more on the cautious side (and strictly uphold the safety requirements) we should add something like this by the cast site:
```rust
if dibv5.as_ptr().align_offset(align_of::<BITMAPV5HEADER>) != 0 {
    return Err(Error::unknown("... This is invalid"));
}
```

It would be nicer to use [is_aligned](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.is_aligned) but that's outside our current MSRV.

wcassels: Cool, I'm happy to do that. Although I guess an alternative - which would save us from having to introduce a new failure mode - is to ptr:;read_unaligned, maybe fiddle, ptr::write_unaligned?

complexspaces: I'm good with that too 👍. We should probably slice off the exact number of bytes the header should be into its own `&mut [u8]` too in order to avoid accidentally clobbering anything.

wcassels: Pushed an initial change (seems to work equally well). 

> We should probably slice off the exact number of bytes the header should be into its own &mut [u8] too in order to avoid accidentally clobbering anything.

Just so I understand fully, are you concerned that we're risking clobbering in the case where `bV5Size < mem::size_of::<BITMAPV5HEADER>()`? 

If so, we could consider reading the size before doing the ptr::read_unaligned, and refuse to do this fiddling if the header is smaller than `mem::size_of::<BITMAPV4HEADER>()` (the earliest documented header with the alpha mask field)

complexspaces: Looking at the implementation you pushed up I think I was just being overly cautious, the logic you wrote is more straightforward then I expected. My concern was that we'd have some offsetting that might subtly break but there's none of that needed.

src/platform/windows.rs
```diff
@@ -198,38 +198,21 @@ mod image_data {
...
-			set(COMPRESSION_OFFSET, BI_BITFIELDS);
-			if red_mask == 0 && green_mask == 0 && blue_mask == 0 {
-				set(MASK_OFFSET, 0xff0000);
-				set(MASK_OFFSET + 4, 0xff00);
-				set(MASK_OFFSET + 8, 0xff);
-			}
-		}
-	}
+        assert!(dibv5.len() >= size_of::<BITMAPV5HEADER>());
+        let src = dibv5 as *mut [u8] as *mut BITMAPV5HEADER;
```
complexspaces: As a suggestion let's explicitly use `as_mut_ptr()` here to avoid needing to think about fat vs thin pointers (it also avoids some `as` casts):
```rust
let src = dibv5.as_mut_ptr().cast::<BITMAPV5HEADER>()
```

wcassels: Done

wcassels:
Also rebased on master to fix the merge conflicts.

complexspaces (approved):
This seems good to me now, thanks again for the PR and I doubly appreciate the added test cases.

~~I was curious if this affected #122 but it seems like its the same regardless of this change. I'll have to continue investigating what's going on there later...~~ EDIT: Nevermind, my brain got two things mixed up here.

wcassels:
Great :confetti_ball: thanks for your help too. Do you usually stick to a regular release schedule? Would be great to move my arboard dependencies back to the crates.io version.

complexspaces:
I don't have a documented schedule for releases. I usually just release once enough has changed or someone asks. Between this and the `file_list` PR I'll probably try and release something Saturday.

complexspaces:

Thank you very much, again, for all the investigation and testing you did before and during this PR, and the collaborative effort. It made landing such a complicated/fragile change a lot easier for me.

Issue #64, opened 2022-07-12
[windows] writing the empty string and reading it back causes panic


jabbalaci:
Under Windows, if I write the empty string on the clipboard (clear it), I expect that if I read it back, I get again the empty string. But in this case, `clipboard.get_text()` panics. The error message is the following:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ContentNotAvailable - "The clipboard contents were not available in the requested format or the clipboard is empty."'
```

The clipboard is empty, true, but shouldn't it return the empty string in this case?

complexspaces:
Hey there @jabbalaci,

Are you unwrapping an error somewhere in your own code, or is this coming from inside `arboard`? I don't think we have any panics in `arboard` which would cause this. 

I believe that the inconsistency here is coming from the crate that we use for clipboard operations. It [skips making any system calls](https://github.com/DoumanAsh/clipboard-win/blob/master/src/raw.rs#L332-L335) when it sees the data is an empty string. It makes sense to do this because otherwise you'd need to make a zero-sized global allocation, etc. While its a little weird, I think it still matches up with the docs for `Error::ContentNotAvailable`.

I'd recommend switching your code to use `.unwrap_or_default()` if you want this empty string behavior.

jabbalaci:
Thanks! I contacted the author of clipboard-win, he made a patch. I think he will publish a new release soon.

complexspaces:
Thanks for bringing that upstream for consideration. Given that we don't pin a specific version of `clipboard-win`, but just a minimum, you should be able to transparently get these changes by running `cargo update -p clipboard-win` in your project. Closing this as done.

PR #203, opened 2025-08-27
Bump wl-clipboard-rs to 0.9.2

sebhoss:
Recent versions of Rust trigger the following warning: `the following packages contain code that will be rejected by a future version of Rust: wl-clipboard-rs v0.9.0`

The fix for that was already done upstream in https://github.com/YaLTeR/wl-clipboard-rs/commit/28b0ca4ce64c9641db1ac8a932569777f675c053 and released in version 0.9.1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ dependencies = [
  "objc2-foundation",
  "parking_lot",
  "percent-encoding",
- "windows-sys",
+ "windows-sys 0.52.0",
  "wl-clipboard-rs",
  "x11rb",
 ]
@@ -123,7 +123,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "778e2ac28f6c47af28e4907f13ffd1e1ddbd400980a9abd7c8df189bf578a5ad"
 dependencies = [
  "libc",
- "windows-sys",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Gae24:
As it is, it's required to bump rust-version to 1.74.0 since there is a usage of [io::Error::other](https://doc.rust-lang.org/std/io/struct.Error.html#method.other), [#64](https://github.com/YaLTeR/wl-clipboard-rs/pull/64) removes it.
`os_pipe` pulls in `windows-sys 0.59`, but there is a [pr](https://github.com/oconnor663/os_pipe.rs/pull/37) to allow more versions of it.

complexspaces:
Hey there, sorry for the silence. I realized I never followed up here.

I would like to wait until my `wl-clipboard-rs` PR merges since it significantly trims down on the dependency churn this would otherwise introduce into `arboard` users' lockfiles.

sebhoss:
No problem really - thanks for maintaining this crate!

atouchet:
wl-clipboard-rs 0.9.3 which includes the aforementioned PR is now out.

complexspaces:
I've merged this update via https://github.com/1Password/arboard/pull/210 with a cleaner lockfile changset. Thanks again for the original PR here!

@Gae24 It looks like `os_pipe` also figured out its `windows-sys` situation, so I figure now was a good time to proceed with this bump. It will also let 1Password cleanup our fork usage.

PR #209, opened 2025-12-24
feat: add optional smithay-clipboard backend for Wayland clipboard support

proletarius101:
- Introduced a new feature `smithay-clipboard` that utilizes the core Wayland `wl_data_device` APIs. This is especially useful to support window compositors like Mutter for GNOME
- Implemented a new `Clipboard` struct in the `smithay` module to handle clipboard operations.
- Added `init_wayland_display` function to initialize the Wayland display pointer for the smithay backend.
- Updated `Cargo.toml` and `Cargo.lock` to include `smithay-clipboard` and its dependencies.
- Created an example `smithay_daemonize.rs` demonstrating the usage of the new backend.
- Enhanced README with details on the limitations and usage of the `smithay-clipboard` feature.
- Updated existing clipboard handling logic to support the new backend while maintaining compatibility with previous implementations.

Related: https://github.com/flathub/me.proton.Pass/issues/37 https://github.com/bitwarden/clients/issues/13431

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,11 @@ dependencies = [
  "objc2-core-foundation",
  "objc2-core-graphics",
  "objc2-foundation",
+ "once_cell",
  "parking_lot",
  "percent-encoding",
- "windows-sys",
+ "smithay-clipboard",
+ "windows-sys 0.52.0",
  "wl-clipboard-rs",
  "x11rb",
 ]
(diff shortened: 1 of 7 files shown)


complexspaces:
Hi there, thanks for the PR.

My understanding of the core Wayland clipboard protocols is that it requires input events to be passed from the compositor into the application receiving those focused user events, which can then perform clipboard operations with the "tokens" available. Does Smithay abstract that away using some internal global state?

PR #213, opened 2026-04-03
Add a `wait_once` to support waiting for just one clipboard request

joshtriplett:
This is useful for programs that put something on the clipboard for single use, and want to exit as soon as it's pasted, rather than waiting for some amount of time.


complexspaces (approved):
Sorry for the delay! I'm happy to merge this once CI is green. It looks like it just needs a quick format.

joshtriplett:
@complexspaces Fixed.

PR #217, opened 2026-04-28
Properly encode/decode MIME=text/uri-list

### LINKED ISSUE #216: Paths Returned by `Get::file_list` have Trailing `\r`
When pasting a file list on X11 (Ubuntu 22.04 LTS) the `PathBuf`s returned by `Get::file_list` have a trailing `\r`.

luca020400:
In all the text/* formats the lines are terminated with a CRLF pair.

Fixes #216


src/platform/linux/mod.rs
complexspaces, line 81: As a minor suggestion, would you mind adding an inline link around here to https://www.rfc-editor.org/rfc/rfc2045#section-2.10 mentioning that CRLF is always generated to match the specification for future proofing?

complexspaces (approved):

complexspaces (approved):

